PR #14, opened 2020-03-19
build: add version with build commit

guggero:
Add a version and the current commit to the built binaries.


version.go
```diff
@@ -0,0 +1,66 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+)
+
+// Commit stores the current commit hash of this build, this should be set using
+// the -ldflags during compilation.
+var Commit string
+
+// semanticAlphabet
```
carlaKC: nit: incomplete comment? Can probably just delete it, pretty self explanatory var.

guggero: Oops, copy paste. Fixed it.

carlaKC: > copy paste

Friend or foe? We will never truly know.

PR #21, opened 2020-04-07
Makefile: add rpc and go.mod check

carlaKC:
rpc-check logic copied across from lnd

Add a check for changes in `go.mod`/`go.sum` by running `go mod tidy` and then grepping for go.mod/ go.sum in the output of `git status`. Chose this approach rather than `git describe --dirty` so that the mod check could be run in between the proto install script and proto check without triggering a false positive on potential rpc changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ DEPGET := cd /tmp && GO111MODULE=on go get -v
 GOBUILD := GO111MODULE=on go build -v
 GOINSTALL := GO111MODULE=on go install -v
 GOTEST := GO111MODULE=on go test -v
+GOMOD := GO111MODULE=on go mod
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -127,6 +124,24 @@ list:
...
+mod:
+	@$(call print, "Tidying modules.")
+	$(GOMOD) tidy
+
+mod-check:
+	@$(call print, "Checking modules.")
+	$(GOMOD) tidy
+	if test -n "$$(git status | grep -e "go.mod\|go.sum")"; then echo "Running go mod tidy changes go.mod/go.sum"; git status; git diff; exit 1; fi
+
+rpc:
```
guggero, line 137: Needs a rebase, otherwise this directive will be added twice.

Issue #23, opened 2020-04-23
Transaction Reports


carlaKC:
Create reports for on-chain and off-chain activity of a node, similar to those produced by [ln-accounting](https://github.com/alexbosworth/ln-accounting). For every off and on chain transaction that occurs within the period specified, generate an accounting entry. Transactions that pay fees will have a separate entry added for the fees paid. 

#### Entry Format
- Timestamp: unix timestamp
- OnChain: whether the transaction was on or off chain
- Amount: expressed in msat for the sake of a consistent unit across on/off chain
- Type: type of transaction, see below
- TxID: transaction id on chain, 
- FromID: source of funds, when available
- ToID: destination of funds, when available
- Fiat: fiat equivalent
- ID: unique identifier, if available
- Note: additional note to provide context

#### Entry Types
Each transaction added to the report should have a type which describes to the reader what lightning related operation the transaction was a part of. The first iteration of these reports will have the following types:
- ChannelOpen: channel opening transaction
- ChannelClose: channel closing transaction
- HtlcForward: fees earned from forwarding off chain
- Receipt: receipt on chain, invoice settle off chain
- Payment: balance spend from wallet on chain, settled payment off chain
- CircularRebalance: special case of off chain send, where destination is our own node, only record fees
- ChannelOpenFee: fees paid to open a channel
- ChannelCloseFee: fees paid to close a channel
- RoutingFee: fees paid off chain for our own payments
- ChainFee: fees pain on chain for our own payments

This classification falls short for on chain transactions, because we cannot distinguish between sends the user initiated on chain and sweeps produced by lnd on chain. This distinction is out of scope for this task. Follow up tasks will work on splitting these categories out further.

#### Export
These reports should be queryable via rpc, or cli interface. The ability to export directly as a CSV should also be supported, as this is a common use case for this kind of report.

Example output:
|Timestamp|On Chain|Amount|Asset|Type|Transaction ID|Fiat|From|To|ID|Note|
|-|-|-|-|-|-|-|-|-|-|-|
[...]

PR #26, opened 2020-04-29
fiat: add historical price estimation using coincap api

carlaKC:
To create accounting reports, we require fiat estimates for BTC prices. This PR adds a fiat package which gets historical prices from coincap's api, and produces price quotes for a set of timestamped BTC values. 

This package is internally required for spreadsheet generation, as in #23, but may be a useful feature for end users as well. For this reason, rpc/cli endpoints are added to expose this functionality (and allow easy testing). 

The coincap api was chosen because it is currently used in our internal spreadsheet generation. The design of this package is intended to allow addition of other data source, should this become a priority. 

One open US question, is how we deal with msat. Currently this PR rounds to the nearest satoshi, since BTC cannot (yet 🤔 ⚡ ) be exchanged for fiat in this unit. However, sub-satoshi amounts are common in ln, so a rounding override which does not round down may be desirable.


frdrpc/rpc.proto
```diff
@@ -294,3 +306,37 @@ message ChannelInsight {
...
+        string id = 1;
+
+        // The amount to be converted to fiat in millisatoshis.
+        uint64 amount_msat = 2;
+
+        // The unix timestamp at which the price should be quoted.
+        int64 timestamp = 3;
+    }
+
+    repeated PriceRequest requests = 1;
```
guggero, line 321: Is there a specific reason to allow multiple requests to be sent at a time? The API is not really intuitive that way with the request ID and the map as a response. But maybe it's needed for a follow-up PR?

carlaKC: This functionality is mostly needed to produce fiat quotes en-masse for an accounting spreadsheet. So it'll be much more efficient to make one large query. 

However, that's going to be done internally in faraday, so rpc could be simplified to just be a single request like the cli? Really just added it for easy testing/ because we might as well have the feature.

guggero: It's okay leaving it that way, was mostly curious. Could be useful for other internal projects to batch multiple requests together.

PR #27, opened 2020-04-29
accounting: produce on chain reports

carlaKC:
Example output:
|Timestamp|OnChain|Type|Amount|Fiat|TxID|Reference|Note|
|-|-|-|-|-|-|-|-|
2020-05-06 14:49:15 +0200 SAST|true|RECEIPT|5000007637000|462974.28|8396a2e749640bb60927a673cf8785b03bb12904444c9d5fb713b9b46e38adcb|8396a2e749640bb60927a673cf8785b03bb12904444c9d5fb713b9b46e38adcb||%!v(MISSING)|%!v(MISSING)
2020-05-06 14:49:15 +0200 SAST|true|LOCAL_CHANNEL_OPEN|-15007637000|-1389.6278|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce|124244814004224|initiated channel with remote peer: 0318946cc61c28b5ed992cd29f7ef7b8f5085d2f45c647b9bcf4232a088ea077bd| capacity: 15000000 sats|%!v(MISSING)|%!v(MISSING)
2020-05-06 14:49:15 +0200 SAST|true|CHANNEL_OPEN_FEE|-7637000|-0.7071458|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce:-1|fees to open channel: 124244814004224|%!v(MISSING)|%!v(MISSING)
2020-05-06 14:49:15 +0200 SAST|true|PAYMENT|-15007637000|-1389.6278|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce||%!v(MISSING)|%!v(MISSING)
2020-05-06 14:49:15 +0200 SAST|true|FEE|-7637000|-0.7071458|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce|dca63c8277f31fa68c34c9fe2def8cf8ee470602fe24dea8b2fa0fb08c7d54ce:-1||%!v(MISSING)|%!v(MISSING)


accounting/docs.md
```diff
@@ -0,0 +1,43 @@
+# Accounting Reports
+Faraday produces accounting reports on a node's on chain and off chain activity. 
+These reports are formatted using the [Harmony Reporting Standard](https://github.com/picksco/harmony). 
+This document provides a description of the entries in these reports. 
+
+## Common Fields
+For brevity, the following fields which have the same meaning for each entry will be omitted: 
+- Timestamp: The timestamp of the block that the channel open transaction appeared in. 
+- Fiat: The value of the amount field in USD. Note that values less than one satoshi will be rounded down to zero.
```
bhandras, line 9: So, say if I only ever transact less then 1 sat transactions (say with zero fee), then the sum would be zero?

accounting/entries.go
```diff
@@ -0,0 +1,123 @@
+package accounting
+
+import (
+	"fmt"
+
+	"github.com/lightningnetwork/lnd/lnrpc"
+)
+
+// getPriceFunc is a function that gets the price of msat value at a given time.
+type getPriceFunc func(amtMsat, ts int64) (float64, error)
```
bhandras, line 10: do we need this declaration twice? (it's in report.go)

accounting/report.go
```diff
@@ -43,5 +43,46 @@ type HarmonyEntry struct {
 	Credit bool
 }
 
+// getPriceFunc is a function that gets the price of msat value at a given time.
+type getPriceFunc func(amtMsat, ts int64) (float64, error)
```
bhandras: I wonder if this will cause a problem in the long run. Accounting may be problematic due to rounding errors (floats can't represent base 10 values exactly [1]).

[1] https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/lightninglabs/loop v0.6.0-beta
 	github.com/lightninglabs/protobuf-hex-display v1.3.3-0.20191212020323-b444784ce75d
 	github.com/lightningnetwork/lnd v0.10.0-beta.rc5
+	github.com/stretchr/testify v1.3.0
```
bhandras, line 12: :1st_place_medal:

accounting/on_chain.go
```diff
@@ -0,0 +1,42 @@
...
+
+		// Our start time is inclusive, skip any transactions that are
+		// strictly before our start time.
+		if timestamp.Before(startTime) {
+			continue
+		}
+
+		// Our end time is exclusive, so we skip any transactions that
+		// are after or equal to our end time.
+		if timestamp.After(endTime) || timestamp.Equal(endTime) {
```
bhandras, line 225: Could just write: `!timestamp.Before(endTime)`

accounting/report.go
carlaKC: Yeah, was also wondering about this. The change would need to happen in the `fiat` package, so will probably follow up on that in a separate PR. Can be achieved by just expressing fiat value in cents with an int. 

A rpc endpoint which returns `fiat_cents` is a bit strange, but probably cleaner in the long run.

accounting/docs.md
carlaKC: As is yes, but would make a change in the fiat package to quote for sub-satoshi values as well. I think the accountant's perspective on this is that we can't exchange these tiny amounts for fiat, so they have an effective 0 value. 

But now that you point it out, maybe better to quote and let people round down themselves? Particularly if we're writing accounting for lightning.

accounting/entries.go
carlaKC: No, accidentally moved it across commits.

accounting/on_chain.go
carlaKC: I knew there was a better way of doing this, brain was clearly on powersaver mode :coffee:

carlaKC:
Thanks for the review @bhandras, definitely some great points about precision for decimal values. 💸 This github.com/shopspring/decimal looks pretty solid, and doesn't have any external deps which is nice. Suggested changes to `fiat` package are in #32.

accounting/docs.md
bhandras: Over time these sub satoshi transactions can add up, so in the end it's better to quote imo.

frdrpc/fiat_estimate.go
```diff
@@ -43,21 +66,11 @@ func parseFiatRequest(req *FiatEstimateRequest) (fiat.Granularity,
 		granularity = fiat.GranularityDay
 
 	default:
-		return granularity, nil, fmt.Errorf("unknown granularity: %v",
-			req.Granularity)
+		return granularity, fmt.Errorf("unknown granularity: %v",
```
bhandras, line 69: nit: early line break

cmd/frcli/node_report.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// If start time is zero, default to a week ago.
+	if req.StartTime == 0 {
+		weekAgo := time.Now().Add(time.Hour * 24 * 7 * -1)
+		req.StartTime = uint64(weekAgo.Unix())
+	}
+
+	// If end time is zero, default to the present.
+	if req.EndTime == 0 {
+		req.EndTime = uint64(time.Now().Unix())
```
bhandras, line 63: nit: this is not needed, as i think this case is also handled on the server side.

accounting/docs.md
```diff
@@ -42,4 +42,16 @@ peers.
...
+Channel close entries represent the on chain close of a channel. 
+
+- Amount: The amount that was paid out to us immediately on channel close. 
+- TxID: The on chain transaction ID for the channel close. 
+- Reference: The channel close transaction ID.
+- Note: A note indicating the type of channel close, and who initiated it. 
+
+Known Omissions: 
+- If our balance is encumbered behind a timelock, or in an unresolved htlc, it will not be paid out as part of this transaction and must be resolved by follow up on chain transactions. 
+- The fees paid to close channels that we initiated are not currently recorded, this is because balances are taken from the funding output rather than being supplied by the wallet.
```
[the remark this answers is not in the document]
carlaKC: For anchors, we actually do pick up fees because we do the sweep back to ourselves to pickup the anchor. This is for legacy channels where fees are just portioned out of your balance.

accounting/docs.md
```diff
@@ -54,4 +54,33 @@ Channel close entries represent the on chain close of a channel.
...
\ No newline at end of file
+- The fees paid to close channels that we initiated are not currently recorded, this is because balances are taken from the funding output rather than being supplied by the wallet.
+
+### Receipt
+A receipt is an on chain transaction which paid to our wallet which was not related to the opening/closing of channels.
+
+- Amount: The amount that was paid to an address controlled by our wallet.
+- TxID: The on chain transaction ID.
+- Reference: The on chain transaction ID.
+- Note: TODO(carla) transaction label once lnd PR is merged.
```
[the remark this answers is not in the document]
carlaKC: Hm, this actually has the weird dependency issue where we need to update `lndClient` in loop to be able to update lnd in faraday. 

I do keep running into this blocker, going to have another look at splitting it out/making it its own module.

accounting/filter_test.go
```diff
@@ -0,0 +1,96 @@
...
+// TestInRange tests filtering of timestamps by a inclusive start time and
+// exclusive end time.
+func TestInRange(t *testing.T) {
+	tests := []struct {
+		name      string
+		timestamp int64
+		inRange   bool
+	}{
+		{
+			name:      "equals start time - ok",
```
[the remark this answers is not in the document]
carlaKC: sgtm!

accounting/entries.go
```diff
@@ -151,3 +151,43 @@ func closedChannelEntries(channel *lnrpc.ChannelCloseSummary,
...
+	txEntry, err := newHarmonyEntry(
+		tx.TimeStamp, amtMsat, entryType, tx.TxHash, tx.TxHash,
+		"", true, convert,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// If we did not pay any fees, we can just return a single entry.
+	if tx.TotalFees == 0 {
```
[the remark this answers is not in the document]
carlaKC: Zero fees in this case actually means that no fees came out of our pocket, but 0 is the nil-equivalent for rpc.

PR #28, opened 2020-05-06
go.mod: update dependencies to lnd v0.10.0-beta

carlaKC:
A few upcoming PRs require a dependency bump, updating now to reduce duplication.


go.mod
```diff
@@ -4,13 +4,13 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/golang/protobuf v1.3.3
-	github.com/grpc-ecosystem/grpc-gateway v1.10.0
+	github.com/grpc-ecosystem/grpc-gateway v1.12.2
```
bhandras, line 7: I wonder why we keep having an old version of grpc-gateway in lnd.

go.mod
```diff
@@ -4,13 +4,13 @@ require (
...
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/lightninglabs/loop v0.2.4-alpha
+	github.com/lightninglabs/loop v0.6.0-beta
 	github.com/lightninglabs/protobuf-hex-display v1.3.3-0.20191212020323-b444784ce75d
-	github.com/lightningnetwork/lnd v0.8.0-beta-rc3.0.20191025122959-1a0ab538d53c
+	github.com/lightningnetwork/lnd v0.10.0-beta.rc5
 	github.com/urfave/cli v1.20.0
-	google.golang.org/genproto v0.0.0-20190819201941-24fa4b261c55
+	google.golang.org/genproto v0.0.0-20190927181202-20e1ac93f88c
```
bhandras, line 13: Same for genproto. We could probably safely upgrade in lnd too.

go.mod
guggero: We can't update `grpc-gateway` in lnd because it has a problem with the `rpc.proto` containing two services. See https://github.com/lightningnetwork/lnd/pull/3650. But that will be fixed with my upcoming proto overhaul PR (lightningnetwork/lnd#4141).

The reason why I would be cautious when updating these in other projects is that if the version used as a dependency and the version used to compile the protos is different, it _could_ potentially lead to problems. Because on a system you only have one set of `protoc` and `genproto` installed (with the versions pinned to what is needed in lnd), you normally don't switch them when compiling these protos. So there could be a difference in theory if the versions drift apart too much.

carlaKC: I think there may be a historical reason for this, but also unsure.

PR #29, opened 2020-05-06
accounting: add harmony report struct

carlaKC:
Add a struct representing a [Harmony](https://github.com/picksco/harmony) accounting record. This struct will form the base of 2 separate PRs (one for off chain accounting, one for on chain), so it is separated out to reduce duplication. I've already written the bulk of the on-chain PR, so I am confident that this struct has the required fields.

### DIFF
--- a/accounting/report.go
+++ b/accounting/report.go
@@ -1,0 +1,47 @@
+package accounting
+
+import (
+	"time"
+
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+// Report contains a set of entries.
+type Report []*HarmonyEntry
+
+// HarmonyEntry represents a single action on our balance.
+type HarmonyEntry struct {
+	// Timestamp is the time at which the event occurred.
+	// On chain events: timestamp will be obtained from the block timestamp.
+	// Off chain events: timestamp will be obtained from lnd's records.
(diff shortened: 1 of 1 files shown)


accounting/report.go
```diff
@@ -0,0 +1,47 @@
...
+	// On chain events: timestamp will be obtained from the block timestamp.
+	// Off chain events: timestamp will be obtained from lnd's records.
+	Timestamp time.Time
+
+	// Amount is the balance change incurred by this entry, expressed in
+	// msat.
+	Amount lnwire.MilliSatoshi
+
+	// Fiat is the fiat value of this entry's amount.
+	Fiat float64
```
bhandras, line 24: nit: Maybe calling this `FiatValue` could work even better.
Q: shouldn't we indicate the currency type as well?

carlaKC: Fiat conversion is handled by another package, which _should_ have multiple fiat conversions supported. To get these reports, you'd query with a currency param. 

But `FiatValue` sgtm!

carlaKC: Fiat conversion is handled by another package, which _should_ have multiple fiat conversions supported. To get these reports, you'd query with a currency param. 

But `FiatValue` sgtm!

PR #32, opened 2020-05-13
fiat: improve precision and allow sub-sat quotes

carlaKC:
Switch the fiat package over from using floats to use `decimal` as defined in github.com/shopspring/decimal which allows for precision up to 15 places. The package does not have any additional dependencies, and provides convenient arithmetic operations. As part of this change, we update the rpc to returns strings - a breaking change, deemed acceptable because we have not included this rpc in a release yet. 

This PR also updates the fiat package to not round quotes to BTC, allowing for sub-satoshi fiat quotes. This is a nice feature to have for lightning based accounting systems; rounding can be done consumer side if required.


fiat/prices.go
```diff
@@ -100,19 +101,21 @@ func getQueryableDuration(requests []*PriceRequest) (time.Time, time.Time) {
 // msatToUSD converts a msat amount to usd. Note that this function coverts
 // values to Bitcoin values, then gets the fiat price for that BTC value. If
 // an amount < 1000 msat is given, a zero amount will be returned.
-func msatToUSD(price float64, amt lnwire.MilliSatoshi) float64 {
-	btcBal := amt.ToBTC()
-	return price * btcBal
+func msatToUSD(price decimal.Decimal, amt lnwire.MilliSatoshi) decimal.Decimal {
+	msatDecimal := decimal.NewFromFloat(amt.ToBTC())
```
bhandras: I wonder if it'd be better to write this as:
```
pricePerMilliSatoshi = price.Div(decimal.NewFromInt(100000000000))
return amt.Mul(pricePerMilliSatoshi)
```

carlaKC: Yeah with a comment this is more readable 👍

bhandras: More like more precise as we avoid the float conversion.

PR #33, opened 2020-05-14
rpcserver+config: prepare to be used as a library

guggero:
To be able to run `faraday` as a library, we need to export some config related elements and create the possibility to instantiate and run the RPC server without it creating its own `grpc.Server` server or REST `proxy.ServeMux` instance.

### DIFF
--- a/cmd/frcli/utils.go
+++ b/cmd/frcli/utils.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -60,23 +60,28 @@ type config struct {
...
 		MinimumMonitored: defaultMinimumMonitor,
 		DebugLevel:       defaultDebugLevel,
 		RPCListen:        defaultRPCListen,
 	}
+}
+
+// LoadConfig starts with a skeleton default config, and reads in user provided
+// configuration from the command line. It does not provide a full set of
+// defaults or validate user input because validation and sensible default
+// setting are performed by the lndclient package.
```
carlaKC, line 78: Just a question about not validating. Once we start adding more values to config here which are not lndclient relevant (something like `minimum_forward_rate` for closing channels that are inactive), where would validation those config settings live?

cmd/frcli/utils.go
```diff
@@ -128,3 +143,42 @@ func clientAddressDialer(defaultPort string) func(context.Context,
 		)
 	}
 }
+
+// readMacaroon tries to read the macaroon file at the specified path and create
+// gRPC dial options from it.
+func readMacaroon(macPath string) grpc.DialOption {
```
carlaKC, line 149: Since we're already importing `lnd/macaroons`, can we add a TODO to extract+export this logic in lnd? Just for the sake of less copied code 🧹

config.go
guggero: Once the need for faraday specific config arises, I think we should probably do it in a similar way as it's done in `lnd` with this PR: https://github.com/lightningnetwork/lnd/pull/4248.
Export a `ValidateConfig` function that can be called after `LoadConfig` to check and sanitize the provided values.

cmd/frcli/utils.go
guggero: Yes, I agree. I didn't want to introduce more dependencies so I copied it for now. But definitely something to keep in mind for a later cleanup. Added a TODO.

PR #35, opened 2020-05-20
accounting: add records for invoices

carlaKC:
Start working on off chain accounting with the addition of invoices. Include the special case where we have paid an invoice to ourselves, as is the case in circular rebalances.


accounting/filter_test.go
```diff
@@ -99,3 +99,33 @@ func TestFilterOnChain(t *testing.T) {
 	expected := []*lnrpc.Transaction{confirmedTx}
 	require.Equal(t, expected, filtered)
 }
+
+// TestFilterInvoices tests filtering out of invoices that are not settled.
+func TestFilterInvoices(t *testing.T) {
+	// Create two invoices within our desired time range, one that is
+	// settled and one that was cancelled.
+	settledInvoice := &lnrpc.Invoice{
```
bhandras, line 107: I'd add a case with a settletime not in range too.

accounting/entries.go
```diff
@@ -191,3 +193,51 @@ func onChainEntries(tx *lnrpc.Transaction,
...
+// overpaid, or both.
+func invoiceNote(memo string, amt, amtPaid int64, keysend bool) string {
+	var notes []string
+
+	if memo != "" {
+		notes = append(notes, fmt.Sprintf("memo: %v", memo))
+	}
+
+	if amt != amtPaid {
+		notes = append(notes, fmt.Sprintf("invoice overpaid "+
```
bhandras, line 207: nit: could be useful to also add the overpay amount.

accounting/entries_test.go
```diff
@@ -388,3 +419,67 @@ func TestOnChainEntry(t *testing.T) {
...
+
+			entry, err := invoiceEntry(
+				invoice, test.circular, mockConvert,
+			)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			expectedEntry := getEntry(test.circular)
+
```
bhandras, line 481: nit: extra new line

accounting/off_chain.go
```diff
@@ -0,0 +1,101 @@
...
+// invoices they paid.
+func offChainReport(invoices []*lnrpc.Invoice, circularPayments map[string]bool,
+	convert msatToFiat) (Report, error) {
+
+	var reports Report
+
+	for _, invoice := range invoices {
+		// If the invoice's payment request is in our set of circular
+		// payments, we know that this payment was made to ourselves.
+		toSelf := circularPayments[invoice.PaymentRequest]
```
bhandras, line 90: nit: this is elegant but will also insert most payments to the map here.

frdrpc/node_report.go
```diff
@@ -5,31 +5,57 @@ import (
 	"fmt"
 
 	"github.com/lightninglabs/faraday/accounting"
+	"github.com/lightningnetwork/lnd/lnrpc"
 )
 
 // parseNodeReportRequest parses a report request and returns the config
-// required to produce a report.
+// required to produce a report containing on chain and off cha.
```
bhandras, line 12: nit: cha -> chain

frdrpc/rpcserver.go
```diff
@@ -140,6 +140,42 @@ func (c *Config) wrapGetChainTransactions(ctx context.Context) func() ([]*lnrpc.
 	}
 }
 
+func (c *Config) wrapListInvoices(ctx context.Context) ([]*lnrpc.Invoice, error) {
+	resp, err := c.LightningClient.ListInvoices(ctx, &lnrpc.ListInvoiceRequest{})
```
bhandras, line 144: nit: should we timeout here?

frdrpc/rpcserver.go
```diff
@@ -140,6 +140,42 @@ func (c *Config) wrapGetChainTransactions(ctx context.Context) func() ([]*lnrpc.
...
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Invoices, nil
+}
+
+func (c *Config) wrapListPayments(ctx context.Context) ([]*lnrpc.Payment, error) {
+	resp, err := c.LightningClient.ListPayments(
```
bhandras, line 153: ** and here

accounting/off_chain.go
carlaKC: How will this insert payments to the map? We populate it in `OffChainReport`, this just checks if our invoice was in a circular payments.

bhandras: I meant it’ll insert items to the `circularPayments` map, not the actual payments. As most are probably not circular it’ll insert upon lookup. Not a big deal just a nit.

frdrpc/rpcserver.go
carlaKC: Added pagination + ability to cancel ctx instead?

frdrpc/rpcserver.go
```diff
@@ -45,6 +45,15 @@ var (
 	// set this to 200MiB atm.
 	maxMsgRecvSize = grpc.MaxCallRecvMsgSize(1 * 1024 * 1024 * 200)
```
bhandras, line 46: Just catched my eye: I guess with pagination we probably don't need this large msg size anymore.

accounting/off_chain.go
bhandras: I think we talked so much about this, that it deserves that two char fix :)

PR #41, opened 2020-05-27
accounting: add off chain payments to node report

carlaKC:
Add payments and circular payments to our accounting report. 
We already account for circular payments on the invoices side, so we mirror that implementation here, making the decision to reflect the amount we send as well as the fees. These entries can cancel each other out, but are included for completeness. Circular payment fees are separated into a different entry type because they are an operational cost which differs from regular fees. 

Some example output:
Timestamp|OnChain|Type|Amount(Msat)|Amount(USD)|TxID|Reference|Note
--|--|--|--|--|--|--|--
2020-05-26 15:22:06 +0200 SAST|false|PAYMENT|-348719000|-31.1548639019555|98ede0bdd200e14b6b298abd11fa28a13f99abddd9fdb2937ef29a6f04eda8cb|17:98ede0bdd200e14b6b298abd11fa28a13f99abddd9fdb2937ef29a6f04eda8cb|Preimage: 59778c9826863b6f438a1b39b07ff9a1e76f7440b2757af7ec5c66957d04a873
2020-05-26 15:22:06 +0200 SAST|false|FEE|-1348|-0.000120431512306|98ede0bdd200e14b6b298abd11fa28a13f99abddd9fdb2937ef29a6f04eda8cb|17:98ede0bdd200e14b6b298abd11fa28a13f99abddd9fdb2937ef29a6f04eda8cb:-1|Settled with: 1 htlc(s)
2020-05-26 15:21:49 +0200 SAST|false|CIRCULAR_PAYMENT|-1337000|-0.1194487625765|e268ad55dc8eda34c585309bbfa2d141419190bcb16eb7d7483d0fdaa85112c3|18:e268ad55dc8eda34c585309bbfa2d141419190bcb16eb7d7483d0fdaa85112c3|Preimage: 06f595b0d4b16bdf9fcbf6aa09c8b4f1a6f25688edef0e9c6ebcab71c73c4128
2020-05-26 15:21:49 +0200 SAST|false|CIRCULAR_FEE|-1001|-0.0000894302253845|e268ad55dc8eda34c585309bbfa2d141419190bcb16eb7d7483d0fdaa85112c3|18:e268ad55dc8eda34c585309bbfa2d141419190bcb16eb7d7483d0fdaa85112c3:-1|Settled with: 1 htlc(s)


accounting/docs.md
```diff
@@ -102,3 +102,33 @@ Circular receipts record instances where we have paid one of our own invoices.
 - TxID: The payment hash of the invoice.
 - Reference: The preimage of the invoice.
 - Note: Optionally set if the invoice had a memo attached, was overpaid, or was a keysend.
+
+### Payment
+Payments off chain represent payments made via the Lightning Network. 
+
+- Amount: The amount that we paid, excluding the off chain fees paid.
```
guggero, line 109: I know, this is pre-existing for the other parts in this document, but it would probably be good to mention this is the amount in millisats, even for on-chain entries.

accounting/entries.go
```diff
@@ -241,3 +241,78 @@ func invoiceEntry(invoice *lnrpc.Invoice, circularReceipt bool,
...
+	return fmt.Sprintf("Preimage: %v", preimage)
+}
+
+// paymentFeeNote creates a note for a payment fee entry.
+func paymentFeeNote(htlcs []*lnrpc.HTLCAttempt) string {
+	return fmt.Sprintf("Settled with: %v htlc(s)", len(htlcs))
+}
+
+// paymentEntry creates an entry for an off chain payment, including fee entries
+// were required.
```
guggero, line 263: nit: `where`?

accounting/off_chain.go
```diff
@@ -97,5 +99,18 @@ func offChainReport(invoices []*lnrpc.Invoice, circularPayments map[string]bool,
 		reports = append(reports, entry)
 	}
 
+	for _, payment := range payments {
```
guggero, line 102: Also pre-existing but found during testing. In the loop above (in `OffChainReport`) we do `cfg.PaidSelf(payment.PaymentRequest)`. That fails for keysend payments where the payment request is empty.

accounting/entries.go
```diff
@@ -241,3 +241,78 @@ func invoiceEntry(invoice *lnrpc.Invoice, circularReceipt bool,
...
+	return fmt.Sprintf("%v:%v", sequenceNumber, paymentHash)
+}
+
+// paymentNote creates a note for payments from our node.
+func paymentNote(preimage string) string {
+	return fmt.Sprintf("Preimage: %v", preimage)
+}
+
+// paymentFeeNote creates a note for a payment fee entry.
+func paymentFeeNote(htlcs []*lnrpc.HTLCAttempt) string {
```
guggero, line 258: Would it make sense to pass in the `payment` itself into these functions so they can select the fields that are relevant for the task at hand?

accounting/off_chain.go
carlaKC: Nice catch, thanks for testing!

accounting/entries.go
carlaKC: I think it makes sense for when we're passing a lot of params, but going to leave as is for these ones. It's a bit easier for testing to be able to pass expected values in.

carlaKC:
> Ran this on my mainnet node and again gained some valuable insights. I probably should pay more with Lightning, had to go way back with the `--start_time` 😂

I volunteer as tribute to receive some stats if you need 🙋😄

Thank you so much for testing out! I'd completely neglected keysend. Addressed this by checking our htlcs on payments (could use payment request, when present, but it's nicer to not need a rpc decode call). Also added a check for duplicate payment hashes, just to simplify things (we can support in future if it ends up being a blocker, but I doubt many people need accounting for the way way back).

accounting/off_chain.go
```diff
@@ -128,3 +135,64 @@ func offChainReport(invoices []*lnrpc.Invoice, circularPayments map[string]bool,
...
+
+		lastHop := hops[len(hops)-1]
+		toSelf := lastHop.PubKey == ourPubkey
+
+		// Before we add our entry to the map, we sanity check that if
+		// it has any duplicates, the value in the map is the same as
+		// the value we are about to add.
+		duplicateToSelf, ok := paymentsToSelf[payment.PaymentHash]
+		if ok {
+			if duplicateToSelf != toSelf {
```
guggero, line 215: nit: could collapse with the previous `if`.

accounting/off_chain.go
```diff
@@ -21,6 +21,13 @@ var (
 	// lnd reflect multiple attempts to pay the same invoice.
 	errDifferentDuplicates = errors.New("duplicate payments paid to " +
 		"different sources")
+
+	// errDuplicatesNotSupported is returned when we see payments with duplicate
+	// payment hashes. This was allowed in legacy versions of lnd, but is not
+	// supported for accounting purposes. Nodes with duplicates will be required to
+	// delete the duplicates or query over a range that excludes them.
+	errDuplicatesNotSupported = errors.New("accounting not supported for " +
+		"legacy nodes with duplicate payment hashes")
```
guggero, line 30: Because users usually don't RTFM, maybe we could add a suggestion to the error message? Like "adjust time range to skip duplicate payments" or something?

PR #42, opened 2020-05-27
accounting: add forwards to node report

carlaKC:
Complete the report saga by adding forwards through our node and the fees we earn. 
While forwards are not balance changing events themselves (not including fees), we include them for completeness + visibility. 

Forwards do not have a unique reference we can use (even if we had payment hash, it is not guaranteed to be unique in a MPP world). We use the incoming and outgoing channel IDs paired with timestamp as a best-effort txid. 

Forwards in lnd are already indexed by timestamp, so we do not need to further filter the entries we retrieve (as was the case for invoices and payments).


accounting/docs.md
```diff
@@ -102,3 +102,26 @@ Circular receipts record instances where we have paid one of our own invoices.
 - TxID: The payment hash of the invoice.
 - Reference: The preimage of the invoice.
 - Note: Optionally set if the invoice had a memo attached, was overpaid, or was a keysend.
+
+### Forwards
+Forward represents a forward through or node. While these entries do not change our balance except for fees earned, we include them for completeness.
```
bhandras, line 107: nit: I'd phrase the beginning to be a bit more specific. Something like: "A forward represents/is a forwarded payment that passes trough one of our channels." (could be wrong though)

accounting/entries.go
```diff
@@ -241,3 +241,52 @@ func invoiceEntry(invoice *lnrpc.Invoice, circularReceipt bool,
 		hash, preimage, note, false, convert,
 	)
 }
+
+// forwardTxid provides a best effort txid using incoming and outgoing channel
+// ID paired with timestamp in an effort to make txid unique per htlc forwarded.
+// This is not used as a reference because we could theoretically have duplicate
+// timestamps.
+func forwardTxid(chanIn, chanOut, unixTx uint64) string {
```
bhandras, line 249: nit/note: If we were to change this to `ts:in:out` then we'd have a natural "global" order which may or may not be useful.

bhandras, line 249: nit: `forwardTxid` could receive a `lnrpc.ForwardingEvent` instead.

accounting/entries.go
```diff
@@ -241,3 +241,52 @@ func invoiceEntry(invoice *lnrpc.Invoice, circularReceipt bool,
...
+func forwardingEntry(forward *lnrpc.ForwardingEvent,
+	convert msatToFiat) ([]*HarmonyEntry, error) {
+
+	txid := forwardTxid(
+		forward.ChanIdIn, forward.ChanIdOut, forward.Timestamp,
+	)
+	note := forwardNote(forward.AmtInMsat, forward.AmtOutMsat)
+
+	fwdEntry, err := newHarmonyEntry(
+		int64(forward.Timestamp), 0, EntryTypeForward, txid, "", note,
```
bhandras, line 271: q: do we need the note? (as chan in/out is in the tx id)

accounting/entries_test.go
```diff
@@ -482,3 +500,45 @@ func TestInvoiceEntry(t *testing.T) {
 		})
 	}
 }
+
+// TestForwardingEntry tests creation of a forwarding and forwarding fee entry.
+func TestForwardingEntry(t *testing.T) {
+	entries, err := forwardingEntry(fwdEntry, mockConvert)
+	if err != nil {
```
bhandras, line 507: You could use `assert.NoError(t, err)` here.

accounting/entries.go
bhandras: Also, should we consider padding the `chanIn/Out` and the `unixTx` to get an equal lenght tx id?

accounting/entries.go
carlaKC: I think it's nice to have the amounts as well? Just so you can eyeball the amounts that you're forwarding.

accounting/docs.md
carlaKC: Definitely better to be more specific, that's what docs are for after all.

PR #47, opened 2020-06-16
ratelimit: put lower boundary on htlc forward interval [PoC]

joostjager:
Proof-of-concept PR to show how htlc interception can be used to rate limit forwards.


joostjager:
Implemented rate limiting with a token bucket in https://github.com/lightningequipment/circuitbreaker

PR #48, opened 2020-06-17
frdrpc+travis: update protobuf version to match lnd

carlaKC:
Once off regenerate of protoc and update travis to check new version.

### DIFF
--- a/scripts/install_travis_proto.sh
+++ b/scripts/install_travis_proto.sh
@@ -5,9 +5,9 @@ set -ev
 
 # Faraday uses the same versions/commits as lnd.
 PROTOC_VERSION=3.4.0
-PROTOBUF_VERSION="b5d812f8a3706043e23a9cd5babf2e5423744d30"
-GENPROTO_VERSION="a8101f21cf983e773d0c1133ebc5424792003214"
-GRPC_GATEWAY_VERSION="v1.8.6"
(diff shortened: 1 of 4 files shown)


scripts/install_travis_proto.sh
```diff
@@ -5,9 +5,9 @@ set -ev
 
 # Faraday uses the same versions/commits as lnd.
 PROTOC_VERSION=3.4.0
-PROTOBUF_VERSION="b5d812f8a3706043e23a9cd5babf2e5423744d30"
-GENPROTO_VERSION="a8101f21cf983e773d0c1133ebc5424792003214"
-GRPC_GATEWAY_VERSION="v1.8.6"
+PROTOBUF_VERSION="v1.3.2"
+GENPROTO_VERSION="20e1ac93f88cf06d2b1defb90b9e9e126c7dfff6"
+GRPC_GATEWAY_VERSION="v1.14.3"
```
guggero, line 10: We should also update the `grpc-gateway` version in `go.mod`.

carlaKC: We already have `github.com/grpc-ecosystem/grpc-gateway v1.14.3` because of our dep on lndclient (merged today).

guggero: Oops, sorry. Was looking at an old state.

PR #50, opened 2020-06-17
accounting: add sweep entry type

carlaKC:
Now that we have switched over to lndclient, we have access to the `WalletKit` rpc which lists our sweeps for us. We can further split out the on chain reporting category by including sweeps and their fees.

```
        {
            "timestamp": "1592395807",
            "on_chain": true,
            "amount": "149000",
            "credit": true,
            "asset": "BTC",
            "type": "SWEEP",
            "txid": "f13e1d9bc18397a67efb96da2da45f0ba25afe4bc592933de5bf3bc644947672",
            "fiat": "0.0141248807824",
            "reference": "f13e1d9bc18397a67efb96da2da45f0ba25afe4bc592933de5bf3bc644947672",
            "note": ""
        },

```

This issue reduces our known omission for on-chain transactions (previously all sweeps were included in here). The only remaining edge case is the success/timeout tx that we broadcast for two stage htlc claim on our own force closes.


accounting/docs.md
```diff
@@ -86,6 +86,27 @@ A fee entry represents the on chain fees we paid for a transaction.
 - Reference: TransactionID:-1. 
 - Note: Note set for fees. 
 
+### Sweep
+A sweep is an on chain transaction which is used to sweep our own funds back to our wallet. This is required when
```
bhandras, line 90: I think you may have accidentally deleted the last part of the sentence after "when".

Issue #49, opened 2020-06-17
frcli nodereport - payment htlc has a route with zero hops


githorray:
```
$ frcli nodereport
[frcli] rpc error: code = Unknown desc = payment htlc has a route with zero hops
```
My node is quite old and has some early payments with no hops stored.

I had a similar problem with joule that was fixed.

https://github.com/joule-labs/joule-extension/pull/207

Versions of faraday and lnd:

frcli version 0.2.0-alpha commit=v0.2.0-alpha-9-gdf39c5a5c9e27d6935d14db1f1c0e7ad2197acc4
lnd version 0.10.99-beta commit=clock/v1.0.0-66-ged93c16e51abfd3512cb9a2800811b598f6768cd

carlaKC:
Thanks for reporting! 

We're currently looking at payment hops to detect whether we have made a payment or ourselves (in the hope of identifying circular rebalances). We needed to look at hops rather than payment request because we don't always have payment request present, but I think it would make sense to fall back to checking the payment request for a destination in this case. If we have neither, I think logging a warning and just accepting that some old circular payments may not be detected is ok. 

In the meantime, you should be able to get around the issue by running `nodereport` with a specific start date that is after this payment.

carlaKC:
> period too long for granularity level

Ah, I know what's going on here. Will get a fix up asap. Basically the api we use to get fiat estimates from has a tradeoff between accuracy of data and the length of time you query for. It's currently hardcoded which isn't good, I'll update it to shift this "granularity" with the length of query. 

>  Unknown desc = payment htlc has a route with zero hops

I am intrigued that you have a payment with no htlcs which has zero hops from the last month. Have you recently upgraded? Also going to put the fix in for this, but I am curious.

githorray:
I have been recklessly running master since launching the node. :)

carlaKC:
If you know which payment is tripping you up, could you find it in `listpayments` for me and post the output? I just want to make sure that the fix I put up covers this case. 

I don't need payment hash/amount or any identifiable fields, so feel free to remove that info. Specifically interested in the payment status, and the list of htlcs (if it's there). 
```
"status": "?",
"htlcs":   [list of htlcs, if present],
```

githorray:
The last payment with an empty set of hops is

```
"SUCCEEDED"
"1531397351000000000"
{
  "total_time_lock": 0,
  "total_fees": "0",
  "total_amt": "X000",
  "hops": [],
[...]

PR #52, opened 2020-06-22
fiat: get best granularity for fiat estimates

carlaKC:
Instead of exposing granularity (the accuracy with which coincap gives us price values) for endpoints that require fiat estimates, add a best granularity function that will adjust granularity based on how long the query period. 

At prices aggregated by day, and with our price logic which splits our queries up into 5 parts, we can query a range of ~100 years before this best granularity fails. So we should be ok for now. 

Partially addresses #49


fiat/coincap_api.go
```diff
@@ -126,6 +131,52 @@ func (g Granularity) minDuration() (time.Duration, error) {
...
+// allows in ascending order so that we can get the best value for a query
+// duration. We require this list because we cannot iterate through maps in
+// order.
+var ascendingGranularity = []Granularity{
+	GranularityMinute, Granularity5Minute, Granularity15Minute,
+	Granularity30Minute, GranularityHour, Granularity6Hour,
+	Granularity12Hour, GranularityDay,
+}
+
+// maxQueryable returns the total amount of time we can query the coincap api
```
bhandras, line 144: nit: `s/maxQueryable/maxSplitDuration`

accounting/conversions.go
```diff
@@ -33,8 +33,13 @@ func invertMsat(msat int64) int64 {
 // getConversion is a helper function which queries coincap for a relevant set
 // of price data and returns a convert function which can be used to get
 // individual price points from this data.
-func getConversion(ctx context.Context, startTime, endTime time.Time,
-	granularity fiat.Granularity) (msatToFiat, error) {
+func getConversion(ctx context.Context, startTime,
+	endTime time.Time) (msatToFiat, error) {
+
```
bhandras, line 38: Should we also check that `!endTime.Before(startTime)`?

fiat/coincap_api.go
```diff
@@ -126,6 +131,52 @@ func (g Granularity) minDuration() (time.Duration, error) {
...
+
+	return maxDuration * maxQueries, nil
+}
+
+// BestGranularity takes a period of time and returns the lowest granularity
+// that we can query the coincap api for taking into account that we allow
+// splitting up of queries into 5 parts to get more accurate price information.
+// If the period of time can't be catered for, we return an error.
+func BestGranularity(duration time.Duration) (Granularity, error) {
+	for _, granularity := range ascendingGranularity {
```
bhandras, line 160: nb/nit: we could simply create `ascendingMaxSplitDuration` const array so we don't calculate `maxSplitDuration` all the time.

fiat/coincap_api.go
```diff
@@ -126,6 +131,52 @@ func (g Granularity) minDuration() (time.Duration, error) {
...
+		// If our target duration is less than the upper limit for this
+		// granularity level, we can use it.
+		if duration <= period {
+			return granularity, nil
+		}
+	}
+
+	// If our duration is longer than all maximum query periods, we fail
+	// and request a query over a shorter period.
+	return GranularityMinute, ErrQueryTooLong
```
bhandras, line 177: Q: why choose `GranularityMinute` instead of `GranularityDay` here?

carlaKC: Really just because I can't return nil without making the function return (*Granularity, error), and pointers to enums are quite abnormal. Replaced with a `""` to make it clearer that it's a nil value.

fiat/coincap_api.go
carlaKC: I'd rather just keep them separate since one is coincap's value and one is our own.

PR #53, opened 2020-06-22
accounting: fallback to payment request inspection for circular payments

### LINKED ISSUE #49: frcli nodereport - payment htlc has a route with zero hops

carlaKC:
To detect payments made to our own node, we call `listPayments` and then match the last htlc in the payment to our own pubkey. This is an issue for older nodes that do not have their htlcs stored on disk. We use our payment request only as a fallback (rather than our main check) because many payments may not have this field set. 

Payments without htlcs:
- In flight payment, initiated but no attempts dispatched
- Legacy payment, complete but no attempts in db

Payments without payment request:
- Keysend payments
- Payments send without `--pay_req` (set dest & amt instead)

Falling back to our payment request covers all cases except legacy payments that have no htlcs set and also do not have a payment request present. In this case, we just log a warning, and document that these payments will not be detected as circular payments (because we cannot access that information). 

Fixes #49 (has been tested for problem payment)


accounting/off_chain.go
```diff
@@ -50,6 +59,9 @@ type OffChainConfig struct {
 	// payments that are made to our own node.
 	OwnPubKey route.Vertex
 
+	// DecodePayReq decodes a payment request.
+	DecodePayReq func(payReq string) (*lndclient.PaymentRequest, error)
```
guggero, line 63: nit: add a function type for this to avoid having the full signature in 3 places?

PR #54, opened 2020-06-22
resolutions: add close report for cooperative close

carlaKC:
This PR is the first part of a series of work which aims to examine the on chain outcome of our channel closes to provide users with a full set of information about channel closes, regardless of whether the information is recorded in lnd (which is often incomplete because we did not previously persist all the information we require). 

This PR adds cooperative close reports, which are relatively simple. Follow ups will produce more complex reports for local and remote force closes. As with accounting, the known omissions (in this case, ability to make reports for batched channel opens) are documented, rather than blocking on PRs to lnd to expose this information.


config.go
```diff
@@ -36,6 +41,9 @@ type Config struct {
 	// Simnet is set to true when using bitcoind's regtest.
 	Regtest bool `long:"regtest" description:"Use regtest"`
 
+	// ChainConn specifies whether to attempt connecting to a bitcoin backend.
+	ChainConn bool `long:"chainconn" description:"Whether to attempt to connect to a backing bitcoin. Some endpoints will not be available if this option is not enabled."`
```
guggero, line 45: nit: s/backing bitcoin/backing bitcoin node/

chain/client.go
```diff
@@ -0,0 +1,135 @@
...
+
+// BitcoinConfig defines exported config options for the connection to the
+// btcd/bitcoind backend.
+type BitcoinConfig struct {
+	Host         string `long:"host" description:"bitcoind/btcd instance address"`
+	User         string `long:"user" description:"bitcoind/btcd user name"`
+	Password     string `long:"password" description:"bitcoind/btcd password"`
+	HTTPPostMode bool   `long:"httppostmode" description:"Use HTTP POST mode? bitcoind only supports this mode"`
+	UseTLS       bool   `long:"usetls" description:"Use TLS to connect? bitcoind only supports non-TLS connections"`
+	TLSPath      string `long:"tlspath" description:"Path to lnd tls certificate"`
```
guggero, line 28: nit: path to btcd TLS certificate (bitcoind doesn't support TLS AFAIK)

chain/client.go
```diff
@@ -0,0 +1,135 @@
...
+
+	tx, err := c.rpcClient.GetRawTransactionVerbose(txHash)
+	if err != nil {
+		return nil, err
+	}
+
+	// Do not cache the transaction if it has not confirmed yet. If we do,
+	// we won't ever lookup the confirmed transaction because it is already
+	// cached.
+	if tx.BlockHash == "" {
```
guggero, line 71: :+1:

chain/client.go
```diff
@@ -0,0 +1,135 @@
...
+// client.
+type BitcoinClient interface {
+	// GetTxDetail looks up a transaction.
+	GetTxDetail(txHash *chainhash.Hash) (*btcjson.TxRawResult, error)
+}
+
+// BitcoinConfig defines exported config options for the connection to the
+// btcd/bitcoind backend.
+type BitcoinConfig struct {
+	Host         string `long:"host" description:"bitcoind/btcd instance address"`
```
guggero, line 23: Must be `host:port`, otherwise it will try port 80.

config.go
guggero: Also, what do you think about `--connect-bitcoin` as an alternative name for the flag?

PR #58, opened 2020-06-28
multi: identify swap spends using script matching and add to accounting

carlaKC:
This PR adds identification of on chain htlc success and timeout transactions associated with subatomic swaps by using script matching to identify spends from htlcs with the same format as lightning lab's loop htlcs. These new types of transactions are added to the existing accounting report in their own category. These transactions will not be confused with on chain htlc sweeps associated with channel closes because we do not use lnd's sweeper to claim swaps on chain.


joostjager:
Or maybe it is possible to do more with tx labels and off-chain payment descriptions. To keep faraday generic. Users may also have their own applications and will need to figure out a way to get that activity in this accounting report too without having the option to special case it in faraday.

carlaKC:
> Or maybe it is possible to do more with tx labels and off-chain payment descriptions. To keep faraday generic. Users may also have their own applications and will need to figure out a way to get that activity in this accounting report too without having the option to special case it in faraday.

Indeed. Would be nice to look at loading your own custom types in config based on labels. 

Closing for now since this isn't required.

Issue #38, opened 2020-05-27
fiat: Expand currency offering


carlaKC:
As is, we only obtain fiat quotes in USD. More options would be nice for non-dollar denominated users. 

[Possible solutions, still in progress]
Investigate whether coincap (currently used price backends) provides other pairs. If not, look at other backends (also relevant for #37) or look at converting BTC -> USD -> Other Currency. Preferably the first option, since each conversion loses accuracy.

carlaKC:
I think that the solution we have in #122 is a more flexible way of allowing multiple currency options. We can sink endless resources into adding new backends/ currencies, so let's rather allow use of the reporting functionality with custom data and re-open this issue if there's demand for additional built in currencies.

PR #60, opened 2020-07-09
fiat: update api to provide timestamped bitcoin prices and use in accounting

carlaKC:
This PR updates the fiat package to return bitcoin prices with timestamps, rather than calculating values from these prices on behalf of calling functions. It also simplifies our price function to just use earlier price datapoints (rather than taking an average). These measures make fiat reporting more auditable, which is a requirements for accounting reports.


fiat/coincap_api.go
```diff
@@ -37,94 +30,94 @@ var (
 
 // Granularity indicates the level of aggregation price information will be
 // provided at.
-type Granularity string
+type Granularity time.Duration
```
joostjager, line 33: Is there a need to granularity at all if we standardize on 1 hour?

accounting/conversions.go
```diff
@@ -67,9 +67,13 @@ func getConversion(ctx context.Context, startTime, endTime time.Time,
...
-			Timestamp: ts,
-		})
+		price, err := fiat.GetPrice(prices, ts)
+		if err != nil {
+			return decimal.Zero, err
+		}
+
+		return fiat.MsatToUSD(
+			price.Price, lnwire.MilliSatoshi(amtMsat),
+		), nil
```
joostjager: No return of the timestamp here to make it traceable?

accounting/conversions.go
```diff
@@ -66,14 +64,7 @@ func getConversion(ctx context.Context, startTime, endTime time.Time,
...
-		price, err := fiat.GetPrice(prices, ts)
-		if err != nil {
-			return decimal.Zero, err
-		}
-
-		return fiat.MsatToUSD(
-			price.Price, lnwire.MilliSatoshi(amtMsat),
-		), nil
+	return func(ts time.Time) (*fiat.USDPrice, error) {
+		return fiat.GetPrice(prices, ts)
```
joostjager, line 63: Looks like something can be squashed if this function is rewritten again in the pr

fiat/coincap_api.go
carlaKC: I think it's nice to be able to give the option of more granular prices. Our requirement is an hour, but if you're generating a weekly report for your node, it's reasonable to want lower.

accounting/conversions.go
carlaKC: This is just to make the commit build. It's not ideal, but combining the accounting change makes for a very large commit.

cmd/frcli/fiat_estimate.go
```diff
@@ -59,10 +56,27 @@ func queryFiatEstimate(ctx *cli.Context) error {
...
+		return fmt.Errorf("expected price for: %v, got: %v", ts,
+			estimate.Timestamp)
 	}
 
+	bitcoinPrice, err := decimal.NewFromString(estimate.BtcPrice.Price)
+	if err != nil {
+		return err
+	}
+
+	usdVal := fiat.MsatToUSD(bitcoinPrice, lnwire.MilliSatoshi(amt))
```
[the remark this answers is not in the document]
carlaKC: To my knowledge, most of the companies interested in reporting (thus far) are US based. There's an open issue #38 for expanding the fiat unit offering, so hopefully people will let us know there if they need other values.

accounting/conversions.go
carlaKC: I understand not wanting to touch code twice in a PR, but sometimes it feels like the cost of the squash (in this case, squashing together changes to completely different packages, making potential future bug hunting more difficult) is more than the benefit of not touching 5 lines twice.

cmd/frcli/fiat_estimate.go
```diff
@@ -44,13 +47,7 @@ func queryFiatEstimate(ctx *cli.Context) error {
 	// Set start and end times from user specified values, defaulting
 	// to zero if they are not set.
 	req := &frdrpc.FiatEstimateRequest{
```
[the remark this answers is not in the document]
carlaKC: Indeed, will rename with next push.

fiat/coincap_api.go
```diff
@@ -37,94 +30,94 @@ var (
...
-	GranularityDay Granularity = "d1"
+	GranularityDay = Granularity(time.Hour * 24)
 )
 
-// maxDuration returns the maximum duration that can be queried for a given
-// level of granularity.
-func (g Granularity) maxDuration() (time.Duration, error) {
+// coincapValue returns the value we need to provide the API with for each level
+// of granularity.
+func (g Granularity) coincapValue() (string, error) {
```
guggero, line 63: If `Granularity` was just a struct with `{Duration time.Duration, CoincapValue string, MaxDuration time.Duration}` these methods could be removed. Would also remove the need for an error return value.
And all values could be in the same place, for example:
`GranularityMinute = newGranularity(time.Minute, "m1", time.Hour * 24)`

carlaKC: Indeed, much nicer this way :)

fiat/coincap_api.go
```diff
@@ -15,121 +15,78 @@ import (
...
+	Granularity12Hour = newGranularity(
+		time.Hour*12, time.Hour*24*365, "h12",
+	)
 
-	default:
-		return 0, errUnknownGranularity
-	}
-}
+	// GranularityDay aggregates the bitcoin price over one day.
+	GranularityDay = newGranularity(
```
guggero, line 86: Nice :sunglasses:

frdrpc/rpc.proto
```diff
@@ -330,25 +330,60 @@ message ChannelInsight {
...
-    map<string, string> fiat_values = 1;
+message BitcoinPrice{
+    // The price of 1 BTC, expressed in USD.
+    string price = 1;
+
+    // The timestamp for this price price provided.
+    uint64 price_timestamp = 2;
+}
+
+message ExchangeRate{
```
guggero, line 377: nit: `make rpc-format`

PR #61, opened 2020-07-20
itests: add setup and node report test

carlaKC:
This PR adds itests to Faraday, starting with coverage of the Node Report endpoint. It is rebased on #63, so the first 5 commits should not be reviewed. 

Our setup is largely copied from the current loop server itests - setups one bitcoin node with two lnd nodes. Instead of starting up a loop client, we start a faraday client.


itest/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN wget https://bitcoin.org/bin/bitcoin-core-${BITCOIND_VERSION}/bitcoin-${BITCOIND_VERSION}-x86_64-linux-gnu.tar.gz && \
+  tar xvfz bitcoin-${BITCOIND_VERSION}-x86_64-linux-gnu.tar.gz
+
+RUN mkdir .bitcoin \
+  && mv bitcoin-${BITCOIND_VERSION}/bin/* /usr/local/bin/
+
+# Get lnd sources and install. Can't use go get here, because tags aren't passed
+# as a compiler variable, leading to lnd not reporting the tags that it is built
+# with.
+RUN git clone https://github.com/lightningnetwork/lnd.git && \
```
guggero, line 30: Would be very useful if we could start creating base images for  the steps up to here and push them to a public repo. That would really speed things up in our itests. Of course not a requirement for this PR, just thinking aloud (and perhaps looking for supporters of this idea :smile: )

run_itest.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
guggero, line 1: Would be nice to add a `make itest` goal that invokes this file.

carlaKC:
> I guess that's because of the force closes?

Indeed, we need to cover sweeping of commit balance for local close so we need to mine a lot. Planning on coming back to this and optimizing, but wanted I wanted to get the coverage in meanwhile.

itest/Dockerfile
carlaKC: Yeah I think that would make a lot of sense!

PR #62, opened 2020-07-23
accounting: get fees for sweeps with using spend notifcations

carlaKC:
wip, want to look into some caching for our spend lookups


carlaKC:
Need pubkey script for spend notifications. Decided to go with a bitcoind connection instead.

PR #63, opened 2020-08-03
accounting: add fees for channel closes and sweeps

carlaKC:
This PR completes the owl for on chain accounting, covering a few holes that were present in the current impl:
- We account for pending channels, so that we do not miss them when a transaction has confirmed on chain, but does not yet reflect as a channel open/close
- Add on chain lookup for sweep and close tx fees, which are not currently reported by lnd's wallet
- Remove double counting of on chain send fees

To get fees for our sweeps and closes, the reporting endpoint requires access to a bitcoin node to lookup the values of our inputs. Once we have addressed this LND side, we can hopefully remove this requirement, but for now we add this requirement so that the accounting report accurately reflects fees.


accounting/entries.go
```diff
@@ -78,22 +53,11 @@ func openChannelFromCloseSummary(channel lndclient.ClosedChannel,
 		entryType = EntryTypeRemoteChannelOpen
 	}
 
-	return openEntries(
-		tx, convert, amtMsat, channel.Capacity, entryType,
-		channel.PubKeyBytes.String(), channel.ChannelID, initiator,
+	ref := fmt.Sprintf("%v", channel.channelID)
```
guggero, line 56: nit: can be just `ref := channel.channelID.String()`

accounting/on_chain.go
```diff
@@ -30,18 +34,85 @@ func OnChainReport(ctx context.Context, cfg *OnChainConfig) (Report, error) {
...
+	// by lnd (this is the case for channel opens that require more than one
+	// conf, or closing channels that are awaiting resolution).
+	pending, err := cfg.PendingChannels()
+	if err != nil {
+		return nil, err
+	}
+
+	// We add our pending force close channels to opened and closed channels
+	// because it is possible that our channel was opened and closed in the
+	// our relevant period.
```
guggero, line 73: nit: extra word `our`.

accounting/on_chain.go
```diff
@@ -102,48 +183,59 @@ func onChainReportWithPrices(cfg *OnChainConfig, getPrice usdPrice) (Report,
...
+	return channelInfo{
+		channelID:    id,
+		channelPoint: chanPoint,
+		pubKeyBytes:  pubkey,
+		capacity:     capacity,
 	}
+}
+
+// onChainReport produces an on chain transaction report.
+func onChainReport(info onChainInformation) (
```
guggero, line 239: This could be passed by reference to avoid copying the data. Or the `onChainReport` could even be a a method of `onChainInformation`.

accounting/on_chain.go
```diff
@@ -178,9 +178,8 @@ func onChainReportWithPrices(cfg *OnChainConfig, getPrice usdPrice) (Report,
 		return nil, err
 	}
 
-	isSweep := make(map[string]bool, len(sweeps))
 	for _, sweep := range sweeps {
-		isSweep[sweep] = true
+		info.sweeps[sweep] = true
```
guggero, line 192: Commit 77b451b759a61fde7aff00f3aac0d2a108132b17 has the itest binaries committed, they should be removed and/or added to gitignore.

resolutions/resolutions.go
```diff
@@ -66,9 +76,26 @@ func ChannelCloseReport(cfg *Config, chanPoint string) (*CloseReport, error) {
...
+
+	// Lookup our transaction and check that it has at most two outputs,
+	// allowing for a funding outpoint and change address. Our current
+	// fee calculations do not account for batched transactions.
+	tx, err := cfg.GetTxDetail(&outpoint.Hash)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(tx.Vout) > 2 {
```
guggero, line 92: Is this condition going to be removed with a future PR? Maybe add a TODO then?

go.mod
```diff
@@ -21,3 +21,5 @@ require (
 )
 
 go 1.13
+
+replace github.com/lightninglabs/lndclient => github.com/carlakc/lndclient v0.0.0-20200730115913-70d01e63b31b
```
halseth, line 25: reminder to update this before merge

accounting/on_chain.go
```diff
@@ -30,18 +34,85 @@ func OnChainReport(ctx context.Context, cfg *OnChainConfig) (Report, error) {
...
+			closeInitiator: lndclient.InitiatorUnrecorded.String(),
+		}
+		info.closedChannels[c.LocalTxid.String()] = closed
+		info.closedChannels[c.RemoteTxid.String()] = closed
+		info.closedChannels[c.RemotePending.String()] = closed
+	}
+
+	// Add our pending open channel to our set of open channels to cover
+	// the case where our transaction has confirmed on chain but has not
+	// yet reached the confirmations lnd requires.
```
halseth, line 110: are you somewhere filtering out non-confirmed channels? The comment makes it sound like they are not included.

fees/fees.go
```diff
@@ -0,0 +1,59 @@
...
+	error) {
+
+	var fees btcutil.Amount
+
+	tx, err := details(txid)
+	if err != nil {
+		return 0, err
+	}
+
+	// First, we minus total outputs from our fees.
```
halseth, line 24: not a big deal, but seems more natural to add inputs first, the subtract outputs?

fees/fees.go
```diff
@@ -0,0 +1,59 @@
...
+		amt, err := btcutil.NewAmount(out.Value)
+		if err != nil {
+			return 0, err
+		}
+
+		fees -= amt
+	}
+
+	// Next, we lookup each of our inputs to figure out their values and
+	// minus them from our fees
```
halseth, line 35: should be adding?

accounting/on_chain.go
```diff
@@ -269,13 +269,24 @@ func onChainReport(info onChainInformation) (
 			continue
 		}
 
-		// Finally, if the transaction is unrelated to channel opens or
-		// closes, we create a generic on chain entry for it. We check
-		// our list of known sweeps for this tx so that we can separate
-		// it our from regular chain sends.
-		isSweep := info.sweeps[txn.TxHash]
+		// Next, we check whether our transaction is a sweep. In this
+		// case we also expect to have zero fees, because sweeps use
```
halseth, line 276: Why will it have zero fees?

accounting/on_chain.go
carlaKC: yes, these opened/closed lists are just build up to match with the output of `listchaintxns`, which we do filter for only confirmed txns. But I see how the comment can be confusing, since that context isn't very clear here, will rephrase!

accounting/on_chain.go
carlaKC: When we use some of the inputs for fees, the wallet doesn't pick up on it in come cases (eg, remote party commit hltcs). Going to fix this in LND so we don't need to do the lookups, this is a workaround. Not always zero though, so updated comment.

carlaKC:
@halseth happy for me to go ahead with merging this, I think I've addressed most of your comments?

halseth:
@carlaKC Yes, looks good! You can possibly update the lnd dep to RC2 before merging :)

PR #68, opened 2020-08-28
A git pull will get accounting and more

AbelLykens:
guggero  07:08
yeah, the go get command changed its meaning in recent go versions. we should probably change our instructions


README.md
```diff
@@ -17,6 +17,7 @@ A [Makefile](https://github.com/lightninglabs/faraday/blob/master/Makefile) is p
 ```
 go get -d github.com/lightninglabs/faraday
```
halseth, line 18: better to just use `git clone` instead? cc @guggero

guggero: Yes, that's what I suggested on Slack as well.

carlaKC:
Fixed by #80.

PR #70, opened 2020-08-31
mod: bump lnd and lndclient versions

guggero:
Update the dependencies to include the final `lnd` release and the tagged `lndclient` version.


carlaKC (approved):
🌮

Issue #69, opened 2020-08-28
OOM error [fatal error: runtime: out of memory], 96 active channels, 1GB


AbelLykens:
I did:
```
go get -d github.com/lightninglabs/faraday
cd $GOPATH/src/github.com/lightninglabs/faraday
git clone
make && make install
```
Set up Faraday to connect to bitcoind backend. Seems to work.

I run faraday. Then in terminal2 I do `frcli insights`. Memory + Swap fills up and faraday crashes. Frcli command tells me then `[frcli] rpc error: code = Unavailable desc = transport is closing`.

I have 96 active channels. Server specs:
```
# free -m
              total        used        free      shared  buff/cache   available
Mem:           1919         522        1135           1         261        1294
Swap:          1639         511        1128
```

@guggero suggested to open an issue ;)

AbelLykens:
```# ./faraday
2020-08-28 07:21:57.681 [INF] FRPC: Starting gRPC listener
2020-08-28 07:21:57.681 [INF] FRPC: gRPC server listening on 127.0.0.1:8465
2020-08-28 07:21:57.681 [INF] FRPC: REST proxy disabled
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0xef7a24, 0x16)
        /usr/local/go/src/runtime/panic.go:774 +0x72
runtime.sysMap(0xc0a0000000, 0x2c000000, 0x181ad98)
        /usr/local/go/src/runtime/mem_linux.go:169 +0xc5
runtime.(*mheap).sysAlloc(0x1802200, 0x2bc60000, 0x0, 0x0)
        /usr/local/go/src/runtime/malloc.go:701 +0x1cd
runtime.(*mheap).grow(0x1802200, 0x15e30, 0xffffffff)
        /usr/local/go/src/runtime/mheap.go:1252 +0x42
runtime.(*mheap).allocSpanLocked(0x1802200, 0x15e30, 0x181ada8, 0x9e)
        /usr/local/go/src/runtime/mheap.go:1163 +0x291
runtime.(*mheap).alloc_m(0x1802200, 0x15e30, 0x7f8bb80a0100, 0x7ffde6a25e98)
        /usr/local/go/src/runtime/mheap.go:1015 +0xc2
runtime.(*mheap).alloc.func1()
        /usr/local/go/src/runtime/mheap.go:1086 +0x4c
[...]

PR #71, opened 2020-08-31
revenue: switchover to use existing pagination function

carlaKC:
Previously, the revenue report used its own pagination function which
returned the wrong offset, resulting in endless querying for forward
events if we used more than one page of events. This change updates
the revenue report to use the existing lndwrap package which has the
correct values.

Fixes: #69

### DIFF
--- a/revenue/revenue.go
+++ b/revenue/revenue.go
@@ -1,16 +1,9 @@
 package revenue
 
 import (
-	"context"
-
-	"github.com/lightninglabs/faraday/paginater"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightningnetwork/lnd/lnwire"
 )
-
-// maxQueryEvents is the number of events to query the forward log for at
-// a time.
-const maxQueryEvents uint64 = 500
 
 // Config contains all the functions required to calculate revenue.
 type Config struct {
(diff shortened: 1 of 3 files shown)


revenue/revenue.go
```diff
@@ -56,91 +47,68 @@ func GetRevenueReport(cfg *Config) (*Report, error) {
...
-			events = append(events, revenueEvent{
-				incomingChannel: incoming,
-				outgoingChannel: outgoing,
-				incomingAmt:     fwd.AmountMsatIn,
-				outgoingAmt:     fwd.AmountMsatOut,
-			})
+			continue
 		}
 
-		return offset, uint64(len(resp.Events)), nil
```
carlaKC, line 131: This was the problem with this function, returning offset rather than resp.LastIndexOffset.

revenue/revenue.go
```diff
@@ -56,91 +47,68 @@ func GetRevenueReport(cfg *Config) (*Report, error) {
...
-// getEvents gets calls the paginated query function until it has all the
-// forwarding events for the period provided. It takes a map of shortChannelIDs
-// to outpoints which is used to convert forwarding events short ids to
-// outpoint strings.
-func getEvents(channelIDs map[lnwire.ShortChannelID]string, queryForwards func(
-	offset, maxEvents uint32) (*lndclient.ForwardingHistoryResponse,
-	error)) ([]revenueEvent, error) {
+// getEvents produces a set of revenue events from a set of forwarding events
+// that can be used to create a report.
+func getEvents(channelIDs map[lnwire.ShortChannelID]string,
```
guggero, line 61: Nice that we can use the code in `lndwrap` to do the actual querying. But now that the meaning of this function has changed, maybe rename it to `toRevenueEvents()` or something to that effect?

guggero (approved):
Nice fix, cool that we can remove some duplicated code.
tACK, tested on my mainnet node.

PR #72, opened 2020-09-08
Add TLS to server connection

guggero:
As a preparation to add macaroon authentication to `faraday`, we first enable TLS on the gRPC and REST server connections.

Because we add a new behavior to the CLI, we might as well streamline the flags a bit and make them more similar to loop in their behavior.


config.go
```diff
@@ -111,6 +123,18 @@ func LoadConfig() (*Config, error) {
 		return nil, fmt.Errorf("error validating network: %v", err)
 	}
 
+	// Clean up and validate paths, then make sure the directories exist.
+	config.FaradayDir = lncfg.CleanAndExpandPath(config.FaradayDir)
+
+	// Append the network type to the data and log directory so they are
+	// "namespaced" per network.
+	config.FaradayDir = filepath.Join(config.FaradayDir, config.Network)
```
carlaKC, line 169: Do we still want to namespace if this is a non-default dir?

config.go
```diff
@@ -135,6 +166,39 @@ func LoadConfig() (*Config, error) {
 		return nil, err
 	}
 
+	// Since our faraday directory overrides our TLS dir values, make sure
+	// that they are not set when loop dir is set. We hard here rather than
```
carlaKC, line 177: s/We hard/We fail hard?

cmd/frcli/utils.go
```diff
@@ -5,14 +5,18 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/lightninglabs/faraday"
```
carlaKC, line 8: nit: import order

cmd/frcli/utils.go
```diff
@@ -128,6 +125,44 @@ func getClientConn(ctx *cli.Context) *grpc.ClientConn {
...
+	// properly read the cert. This will either be the default, or will have
+	// been overwritten by the end user.
+	faradayDir := lncfg.CleanAndExpandPath(ctx.GlobalString(
+		faradayDirFlag.Name,
+	))
+	tlsCertPath := lncfg.CleanAndExpandPath(ctx.GlobalString(
+		tlsCertFlag.Name,
+	))
+
+	// If a custom lnd directory was set, we'll also check if custom paths
```
carlaKC, line 150: nit: lnd-> faraday

carlaKC (commented):
Familiar set of changes, just a question about namespacing `faradaydir` by default?

config.go
guggero: I don't have a strong opinion on it. My thought was to be as similar as possible to loop so users can expect the same behavior. But it would make sense to change it to non-namespaced directories in both projects?

cmd/frcli/utils.go
guggero: Oops :see_no_evil:.
Literally the only thing that annoys me in GoLand :joy: If they fix the auto import placement, this is the perfect Golang IDE for me :)

guggero:
Fixed the itest and some linter issues as well.

config.go
```diff
@@ -111,6 +123,18 @@ func LoadConfig() (*Config, error) {
 		return nil, fmt.Errorf("error validating network: %v", err)
 	}
 
+	// Clean up and validate paths, then make sure the directories exist.
+	config.FaradayDir = lncfg.CleanAndExpandPath(config.FaradayDir)
+
+	// Append the network type to the data and log directory so they are
```
carlaKC, line 167: s/to the data and log directory/to faraday directory

config.go
carlaKC: No let's leave it as is and keep it consistent across the two.

carlaKC (approved):
LGTM!

PR #73, opened 2020-09-08
itest: speed up image build

guggero:
With this PR we first speed up the build by removing the need to download bitcoind every time we change something in our itest folder (this should shave off around 2 minutes of our current itest run). This is achieved by splitting up the docker build into two parts, each with its own base image. For the actual execution we use the ruimarinho/bitcoin-core base image that already contains the bitcoind binary. We use the same image in our testnet cluster environment so it should be fine to use for the integration tests.

We also update to the final version of `lnd v0.11.0` and fix the itest log upload.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,5 +37,5 @@ jobs:
       name: Test integraton
 
 after_script:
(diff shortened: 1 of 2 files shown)


carlaKC (approved):
LGTM!

PR #74, opened 2020-09-08
makefile: add import order check

carlaKC:
This PR adds an import ordering check which can be used to detect stray newlines and mis-grouped imports using https://github.com/pavius/impi. This is chosen over `goimports`, because go imports run for everything in your `GOPATH`, which can be very slow locally. 

This tool cannot be applied from a start commit hash, so requires a once-off fix of import ordering in the project, which is included. This was done manually, because the tool only logs output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,14 @@ LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
 GOVERALLS_PKG := github.com/mattn/goveralls
 GOACC_PKG := github.com/ory/go-acc
 
+IMPORTS_PKG := github.com/pavius/impi/cmd/impi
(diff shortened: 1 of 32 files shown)


Makefile
```diff
@@ -118,6 +126,11 @@ fmt:
 	@$(call print, "Formatting source.")
 	gofmt -l -w -s $(GOFILES_NOVENDOR)
 
+imports: $(IMPORTS_BIN)
+	@$(call print, "Checking imports.")
+	$(IMPORTS_BIN)  --local github.com/lightninglabs/faraday/ --scheme stdThirdPartyLocal --ignore-generated=true *
```
guggero, line 131: nit: extra space before first argument.

guggero (approved):
Nice, very useful tool! LGTM :100:

Issue #67, opened 2020-08-24
unable to run node report


githorray:
```
$ faraday --version
faraday version 0.2.0-alpha commit=v0.2.0-alpha-69-g5bb8bc008fdbaecfec43e76bab117a75ca2588a1

$ lnd --version
lnd version 0.11.0-beta commit=v0.11.0-beta-30-gc3821e5ad19e374ef2454cc9c7bf81b74408b8ca

$ cat ~/.lnd/lnd.conf | grep bitcoin.node
bitcoin.node=neutrino

$ frcli nodereport
[frcli] rpc error: code = Unknown desc = bitcoin node required
```

carlaKC:
The README has _just_ been updated to document this [new requirement](https://github.com/lightninglabs/faraday#chain-backend). 

The `NodeReport` endpoint requires a connection to a `bitcoind` or `btcd` node to fully report on on chain fees. This is required for sweeps and channel closes where the fees for the transaction are taken from existing inputs which lnd's wallet does not know belong to us. Fees report as 0, when we've actually used some of our input as fees. For reliable accounting, we need to account for these fees, so we require a bitcoin node to lookup the values of these inputs and report on their fees. 

Is connecting faraday to a bitcoin node going to be an issue for you? I've been considering adding a flag to the report `--nonode` which allows you to generate best-effort reports, accepting that they will not have fees reported for sweeps/channel closes. Ideally, we will fix this in lnd's wallet and then remove the need for a node, but this could be a workaround in the meantime?

githorray:
I saw the readme changes when tracking down the cause of the error after the most recent commits. 

It would be nice to be able to run the command via `--nonode` or whatever solution is best for neutrino nodes.

githorray:
Updated to the latest commit. Getting a new error:

```
$ frcli --version
frcli version 0.2.0-alpha commit=v0.2.0-alpha-90-g2415b5d866838880f7ddc0f57a79e3c9e589dbda
$ frcli nodereport
[frcli] open : no such file or directory
```

carlaKC:
> [frcli] open : no such file or directory

Are you running on testnet/regtest? We've recently added tls, so you'll need to specify `--network=regtest/testnet` when using `frcli`. But this should have a better error (since the addition of TLS is new) so I will update that). 

If you're running on mainnet however, it should just work.

carlaKC:
[...]

PR #75, opened 2020-09-09
frcli: wrap tls cert error with explanation

carlaKC:
With the addition of TLS, the `network` flag is now required to use `frcli` on regtest/testnet. 
Running [without the flag yields](https://github.com/lightninglabs/faraday/issues/67#issuecomment-689232160) `[frcli] open : no such file or directory`, which is quite ambiguous.

This PR wraps the error to more clearly explain what's gone wrong. 
Mentioned in #67


guggero (approved):
LGTM, thanks for the fix :1st_place_medal:

PR #77, opened 2020-09-09
Add macaroon authentication

guggero:
Next we add macaroon authentication to the RPC to be able to restrict access to authorized users only.
With these two features enabled by default, the loop RPC port can be opened to the internet because unauthorized access is no longer possible.

By default only one macaroon is created for faraday and stored in a file called ~/.faraday/<network>/faraday.macaroon


frdrpc/macaroons.go
```diff
@@ -0,0 +1,160 @@
...
+	// create/unlock calls in the RPC. Using a password should be optional
+	// though.
+	macDbDefaultPw = []byte("")
+)
+
+// startMacaroonService starts the macaroon validation service, creates or
+// unlocks the macaroon database and creates the default macaroon if it doesn't
+// exist yet. If macaroons are disabled in general in the configuration, none of
+// these actions are taken.
+func (s *RPCServer) startMacaroonService() error {
```
carlaKC, line 92: is "if macaroons are disabled.." relevant here since we don't have a `--no-macaroons` option?

frdrpc/rpcserver.go
```diff
@@ -148,10 +156,28 @@ func (s *RPCServer) Start() error {
...
+	serverOpts := s.macaroonInterceptor()
+	s.grpcServer = grpc.NewServer(serverOpts...)
+
 	// Start the gRPC RPCServer listening for HTTP/2 connections.
 	log.Info("Starting gRPC listener")
 	grpcListener, err := net.Listen("tcp", s.cfg.RPCListen)
 	if err != nil {
+		// The macaroon service is the only thing we started yet, so
+		// clean that up now before we exit.
+		if err := s.stopMacaroonService(); err != nil {
```
carlaKC, line 177: Not sure whether this is more simple, but we could have:
```
shutdownFuncs []func()error
defer func(){
  s:= range shutdownFuncs; s()
}
```

Then just append as we go and set the slice to nil if we startup successfully.

README.md
```diff
@@ -32,17 +32,25 @@ Faraday connects to a single instance of lnd. It requires access to macaroons fo
...
-If no custom loop directory is set then the TLS certificate is stored in
-`~/.faraday/<network>/tls.cert`.
+If no custom faraday directory is set then the TLS certificate is stored in
+`~/.faraday/<network>/tls.cert` and the base macaroon in
+`~/.faraday/<network>/faraday.macaroon`.
 
-The `frcli` command will pick up the file automatically on mainnet if no custom
-loop directory is used. For other networks it should be sufficient to add the
-`--network` flag to tell the CLI in what sub directory to look for the files.
+The `frcli` command will pick up these file automatically on mainnet if no
```
carlaKC, line 44: nice 👌

carlaKC (approved):
LGTM, just nits 🌵

config.go
```diff
@@ -133,31 +131,13 @@ func DefaultConfig() Config {
...
-	appName := filepath.Base(os.Args[0])
-	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
-	if config.ShowVersion {
-		fmt.Println(appName, "version", Version())
-		os.Exit(0)
-	}
-
+// ValidateConfig sanitizes all file system paths and makes sure no incompatible
+// configuration combinations are used.
+func ValidateConfig(config *Config) error {
```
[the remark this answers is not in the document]
guggero: Yes, exactly. We load one big config file that contains all options for lnd, loop and faraday. But we validate each section individually by using the `Validate` functions in the main packages.

frdrpc/macaroons.go
guggero: Right, that was left over from when we still had the `--no-macaroons` in the loop PR.

frdrpc/rpcserver.go
guggero: I like that idea, done.

config.go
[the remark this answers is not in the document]
carlaKC: Thought so 👍

carlaKC (approved):
Had a look at the latest diff, looks good!

PR #83, opened 2020-09-18
mod+macaroons: bump to lnd v0.11.1-beta.rc3

guggero:
To make sure we can use the latest version of faraday in LiT, we need to
make sure faraday compiles against the actual v0.11.1 branch of lnd. There
were some conflicting changes to the macaroon service that didn't make
it to the branch that we need to roll back. Those changes don't affect
faraday at all as we were always using the default root key ID anyway.


carlaKC (approved):
LGTM!

PR #85, opened 2020-10-08
audit: add custom categories to reporting

carlaKC:
This PR adds the ability for users to specify a named set of regexes (called a category) which add the ability to customize the output of the `NodeAudit` endpoint. 

For example, if you wanted to group all of your loop transactions in a category called `loop transactions`, you could run the following:
`frcli --network=regtest audit --categories="loop transactions/regex/loopd --/regex/swap"`

This would identify all of the on chain transactions and off chain receives related to loop. Unfortunately, this would not include off-chain payments, because  they are not currently labelled by lnd. 

Two possible approaches here: 
1. Keep our existing entry type enum as is and supplement with a category name
- Advantage: you still know what type a transaction was (sweep, channel, fee etc) within a custom group
- Disadvantage: when sorting by entry type, the custom categories still need to be separated out by category name

|Timestamp|OnChain|Type|Category|Amount(Msat)|Amount(USD)|TxID|Reference|BTCPrice|BTCTimestamp|Note|
|-|-|-|-|-|-|-|-|-|-|-|
10:16:38|true|FEE|loop transactions|5000000000000|0|58482e73b8ac317cafe38d421064ece781e9a4d4de24002368524c2a98d95af4|58482e73b8ac317cafe38d421064ece781e9a4d4de24002368524c2a98d95af4|0|0|swap
10:17:08|true|RECEIPT|loop transactions|100000000000|0|19084faf3ff1e92d3a55f4772057b220bd5a205ba22d7c5ad82d3bd344432b37|19084faf3ff1e92d3a55f4772057b220bd5a205ba22d7c5ad82d3bd344432b37|0|0|swap
10:18:59|false|PAYMENT|loop transactions|5000013120000|0|7b0994640faedda167c99c5696cdcf54235fd1d280d3add60637534c3ad0b941|7b0994640faedda167c99c5696cdcf54235fd1d280d3add60637534c3ad0b941|0|0|lsat

2. Add `CUSTOM` to the entry type enum, and supplement with the category name in a new field:
  - Advantage: do not need to sort custom categories out of all the existing entry types
  - Disadvantage: we loose the entry type 

|Timestamp|OnChain|Type|Category|Amount(Msat)|Amount(USD)|TxID|Reference|BTCPrice|BTCTimestamp|Note|
|-|-|-|-|-|-|-|-|-|-|-|
 10:16:38|true|CUSTOM|loop transactions|5000000000000|0|58482e73b8ac317cafe38d421064ece781e9a4d4de24002368524c2a98d95af4|58482e73b8ac317cafe38d421064ece781e9a4d4de24002368524c2a98d95af4|0|0|swap
10:17:08|true|CUSTOM|loop transactions|100000000000|0|19084faf3ff1e92d3a55f4772057b220bd5a205ba22d7c5ad82d3bd344432b37|19084faf3ff1e92d3a55f4772057b220bd5a205ba22d7c5ad82d3bd344432b37|0|0|swap
10:18:59|false|CUSTOM|loop transactions|5000013120000|0|7b0994640faedda167c99c5696cdcf54235fd1d280d3add60637534c3ad0b941|7b0994640faedda167c99c5696cdcf54235fd1d280d3add60637534c3ad0b941|0|0|lsat

At present I've gone with 1, since I see the use case being you set your regex to catch loop transactions then still want to identify the fees you pay, but open to suggestions here.


cmd/frcli/csv.go
```diff
@@ -21,8 +21,8 @@ func csv(e *frdrpc.ReportEntry) string {
 
 	ts := time.Unix(int64(e.Timestamp), 0)
 
-	return fmt.Sprintf("%v,%v,%v,%v%v,%v%v,%v,%v,%v,%v,%v",
-		ts, e.OnChain, e.Type, amountPrefix, e.Amount, amountPrefix,
-		e.Fiat, e.Txid, e.Reference, e.BtcPrice.Price,
+	return fmt.Sprintf("%v,%v,%v,%v,%v%v,%v%v,%v,%v,%v,%v,%v",
```
guggero, line 24: Missing comma after value!

accounting/categories.go
```diff
@@ -14,6 +16,25 @@ type CustomCategory struct {
 	Regexes []*regexp.Regexp
 }
 
+// NewCustomCategory creates compiles the set of regexes provided and returning
+// a new category. This function assumes that each regex is unique.
+func NewCustomCategory(name string, regexes []string) (*CustomCategory, error) {
```
guggero, line 21: Is there a specific reason why a pointer is returned here while the slice always contains non-pointer elements? IMO we should always use pointers to structs. Or is the main goal to avoid nil references?

cmd/frcli/node_audit.go
```diff
@@ -13,10 +13,40 @@ import (
...
+	regular delimiters):
+	name:%vregularexpression%vregularexpression
+
+	To specify a single category provide the following: 
+	--categories=name:%vregularexpression}%vregularexpression
+
+	Multiple categories need to be wrapped in braces and comma 
+	separated, but may *not* contain any spaces
+	--categories={categorystring,categorystring}
+`, splitter, splitter, splitter, splitter),
```
guggero, line 49: nit: can use `%[1]v` to always use the the same argument (the number in the bracket is the one-based index).

cmd/frcli/node_audit.go
```diff
@@ -13,10 +13,40 @@ import (
...
+	the labels for on chain transactions and the invoices on memos 
+	(at present we cannot match forwarding events and payments).
+
+	Each category is specified as follows, using /regex/ to delimit
+	sections (because regular expressions themselves contain most
+	regular delimiters):
+	name:%vregularexpression%vregularexpression
+
+	To specify a single category provide the following: 
+	--categories=name:%vregularexpression}%vregularexpression
```
guggero, line 44: nit: stray `}`.

cmd/frcli/node_audit.go
```diff
@@ -41,6 +71,12 @@ var onChainReportCommand = cli.Command{
 			Name:  "enable_fiat",
 			Usage: "Create a report with fiat conversions.",
 		},
+		cli.StringSliceFlag{
```
guggero, line 74: What do you think of using JSON instead? We do that in `lnd` in some places.
You could then just specify:
```
--categories='{"name":["regex1","regex2"],"name2":["regex"]}'
```
Or if you want to allow overwriting the boolean flags as well (assuming they're true by default):
```
--categories='{"name":{"patterns":["regex1","regex2"],"offChain":false},"name2":{"patterns": ["regex"]}}'
``

File: cmd/frcli/node_audit.go [hunk 6]
```diff
@@ -41,6 +71,12 @@ var onChainReportCommand = cli.Command{
...
+			Name: "categories",
+			Usage: "A set of custom categories to create the " +
+				"report with, interpreted as category1:" +
+				"regex,regex",
+		},
```
guggero, line 79: Would be very nice if there was a flag like "use-default-categories" that then uses a pre-defined set of patterns that contains what the labels look like in loop and other projects. That's what most people would want to use anyway, right?

guggero (commented):
Very nice addition, this will allow for some awesome RegExp magic :mage_woman: 

Just some comments around the CLI, otherwise looks very good!

accounting/categories.go
carlaKC: I used a pointer here mostly because it's convenient to be able to return nil values on errors, and as you say it generally makes sense to use pointers for structs.

As for the slice, I'm trying to get out of the habit of slices of pointers generally: https://philpearl.github.io/post/bad_go_slice_of_pointers/. I think that we discussed this a while back as a team, or in some channel, but never really acted on it. I think that it makes sense to not use slices of pointers as a rule, but it does have a bit of friction with pointers-everywhere-for-structs, because it seems a bit unintuitive as you point out.

cmd/frcli/csv.go
carlaKC: This is actually intentional, because the double `%v%v` is for an `amountPrefix` and `amount` to show whether amounts are negative. Should really use a library for this rather than rollin our own csv, on the to do list 😬

cmd/frcli/node_audit.go
carlaKC: TIL!

cmd/frcli/node_audit.go
carlaKC: Great idea! Going to do it on a per-project basis, so I have added one for loop now. Can't wait to expand the list 😜

cmd/frcli/node_audit.go
carlaKC: Yeah this is definitely better! Updated to allow json input, decided to just go with the rpc struct directly since it's already annotated.

cmd/frcli/csv.go
guggero: Oops, I missed that, sorry :sweat_smile:

accounting/categories.go
guggero: Thanks for the article, that makes a lot of sense. I guess the reason why I find slices of structs weird is because you cannot tell the difference between a member that happens to use all default values vs. a member that wasn't initialized at all.
With pointers you would at least run into a nil reference error instead of just silently using the default values. But that probably almost never leads to issues if you're just looping over the whole slice anyway.

cmd/frcli/node_audit.go
guggero: Nice!

accounting/categories.go
```diff
@@ -1,6 +1,8 @@
 package accounting
 
-import "regexp"
+import (
+	"regexp"
```
guggero, line 4: nit: should be added to previous commit.

cmd/frcli/node_audit.go
```diff
@@ -17,6 +18,36 @@ var onChainReportCommand = cli.Command{
...
+	the labels for on chain transactions and the invoices on memos 
+	(at present we cannot match forwarding events and payments).
+
+	Categories should be expressed as a json array with the 
+	following format:
+	--categories='[
+		{ 
+			"name": "test", 
+			"on_chain": true, 
+			"label_identifiers": ["swap", "test"]
```
guggero, line 43: Nice that we can use the RPC struct directly! Maybe we should rename the `label_identifiers` field to something like `label_patterns` or `label_imatchers` though to make it more apparent that you can specify regular expressions (in this example specifically where you don't see the RPC docs)? Or just add an example pattern that looks more like regex?

cmd/frcli/node_audit.go
```diff
@@ -114,6 +122,19 @@ func queryOnChainReport(ctx *cli.Context) error {
 		req.CustomCategories = categories
 	}
 
+	if ctx.Bool("loop-category") {
+		req.CustomCategories = append(req.CustomCategories,
```
guggero, line 126: nit: formatting.

guggero (approved):
Great work, LGTM! Only nits remaining :taco:

cmd/frcli/node_audit.go
carlaKC: Renamed and added a more obviously regexey example!

accounting/categories.go
carlaKC: Yeah, I go back and forth on it myself. Perhaps worth a more general discussion re what we want to do across projects.

PR #88, opened 2020-10-29
config: use single macaroon

### LINKED ISSUE #84: Don't require all macaroons to use Faraday
Currently users are required to specify a directory for LND's macaroons with `--lnd.macaroondir`. This can be problematic or annoying if you aren't running Faraday on the same server that's running LND. I would suggest the ability to use a single macaroon (like `admin.macaroon`) to make it easy to use. Eventually once the macaroon baking changes land in LND we could even bake a macaroon specific for Faraday.

carlaKC:
Fixes #84


carlaKC:
approach will be different with final version of custom macaroon, closing.

PR #89, opened 2020-10-30
mod+itest: update lnd and lndclient deps

guggero:
To get rid of a `replace` directive in LiT, we want to update to the latest `lndclient` and `lnd` versions.

This does not add any new runtime version dependencies as RPC level compatibility isn't affected.


carlaKC (approved):
LGTM!

PR #91, opened 2020-11-17
Remove duplicate msat unit in outputs

C-Otto:
Before: "incoming: 51 mSAT msat outgoing: 51 mSAT msat"
After:  "incoming: 51 mSAT outgoing: 51 mSAT"


carlaKC (approved):
LGTM, thanks for the fix!

PR #98, opened 2020-12-02
multi: add pool category and in-repo release notes

carlaKC:
This PR combines two small changes for brief review:
1. Now that pool has been released, we can add a pre-set custom category for pool so that pool txns can easily be separated out from regular lnd txns
2. Add in-repo release note tracking, and reminders in comments + PR template


guggero (approved):
Very nice, LGTM :100:

PR #100, opened 2021-01-04
go.mod: update to latest lndclient

carlaKC:
Contains bugfix in github.com/lightninglabs/lndclient/pull/33 which
we need for our revenue report.


bhandras (approved):
LGTM

PR #110, opened 2021-03-29
frdrpc: rename proto, use Docker for compilation, move to GitHub CI

guggero:
The main purpose of the PR is to rename `rpc.proto` to `faraday.proto` in an attempt to get rid of the error `WARNING: proto: file "rpc.proto" is already registered` we see wherever we use both Faraday as well as `lnd` as libraries.

While we're at it, we also use Docker for compiling the RPC protos, update the golang version and move the CI pipeline to GitHub.


frdrpc/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  protobuf-compiler='3.6.1*' \
+  clang-format='1:7.0*'
+
+# We don't want any default values for these variables to make sure they're
+# explicitly provided by parsing the go.mod file. Otherwise we might forget to
+# update them here if we bump the versions.
+ARG PROTOC_GEN_VERSION
+ARG GRPC_GATEWAY_VERSION
+
+ENV FALAFEL_VERSION="v0.7.1"
```
carlaKC, line 14: `FALAFEL_VERSION` unnecessary?

go.mod
```diff
@@ -21,4 +21,4 @@ require (
 	gopkg.in/macaroon.v2 v2.1.0
 )
 
-go 1.13
+go 1.15
```
carlaKC, line 24: nit: commit message doesn't quite fit, since we're bumping min version and docker versions to different values

carlaKC (approved):
Great improvement! Just some nits.

frdrpc/Dockerfile
guggero: Correct, obviously this isn't needed :sweat_smile:

PR #115, opened 2021-04-20
fiat: expand currency offering

### LINKED ISSUE #38: fiat: Expand currency offering

ellemouton:
Allows a user to choose a currency other than USD for the `audit` and `fiat` RPCs. CoinCap's API is used to fetch a list of available usd rates and if the wanted currency is found then all results are converted to that currency using the given usd Rate. If no currency flag is provided then the default is still USD.

Fixes #38 

For example:

```
$ frcli fiat --amt_msat=100000000000 --currency=ZAR

100000000000 msat = 803341.50837 ZAR, priced at 2021-04-20 12:17:00 +0200 SAST
```


fiat/coincap_api.go
```diff
@@ -274,3 +325,47 @@ func (c *coinCapAPI) GetPrices(ctx context.Context, startTime,
...
 }
+
+// coinCapRatesResponse holds the list of conversion rates returned by CoinCap.
+type coinCapRatesResponse struct {
+	Data []*coinCapRatePoint `json:"data"`
+}
+
+// coinCapRatePoint holds a currency symbol along with its conversion
+// rate to USD.
+type coinCapRatePoint struct {
```
carlaKC, line 336: Say we're after the BTC/ZAR price. If we find our USD/ZAR rate in this response, then use it to convert our USD/BTC prices, wouldn't it be a case that a bitcoin price from last week will be converted to ZAR using today's price?

carlaKC (commented):
Original thinking behind this was that we'd query the BTC/currency directly rather than offer conversion. However that does limit us to currencies that have direct listings at price endpoints. 

Not against getting USD/BTC and converting, but I think that it's important to offer fiat price at time of transaction (which our USD/BTC data points currently offer). 

Questions that seem relevant to me here are:
- How many currency pairs can we get direct BTC prices for?
- What are our options for timestamped USD currency pairs if we can't get a lot of listings. 

I'll do some googling 🇿🇦🤘

Issue #37, opened 2020-05-27
fiat: Expand fiat backends


carlaKC:
As is, the fiat package only uses CoinCap's API for price information. The package is designed to allow different price sources without affecting our price calculation logic. 

### Steps to completion
1. Identify different sources of price data and propose in this issue (perhaps [coindesk](https://www.coindesk.com/coindesk-api))
2. Add code to obtain price data over a range and update `fiat/GetPrices` to allow users to choose a backend
3. Update rpc to allow choice of backends

nixbitcoin:
It would also be great to create a Proxy option, like `--server.proxy` in loopd, to tunnel all price requests over Tor. This has obvious privacy benefits.

@zkSNACKs offer a Tor v3 HS under which you can requests prices at `http://wasabiukrxmkdgve5kynjztuovbg43uxcbcxn6y2okcrsg7gb6jdmbad.onion/api/v3/btc/Offchain/exchange-rates`.

ellemouton:
I was having a go at adding the `--server.proxy` option that @nixbitcoin suggested. However, it seems to me that CoinCap does not allow requests over Tor. When I curl the API directly I get json back but when I curl the API through the Tor proxy then I get a webpage back that requires a captcha to be completed. The same does not go for other APIs such as coindesk which makes me think that CoinCap is not allowing Tor requests (i am a Tor noob though so there may be a way to get around this that I am unaware of).

PR #116, opened 2021-04-20
multi: add Coindesk backends for fiat prices

ellemouton:
This PR addresses #37 by adding CoinDesk as a possible backend to use for price information. Users can use the `fiat_backend` flag in the `audit` and `fiat` RPCs to chose a backend. The default is still CoinCap.

For example:

```
$ frcli fiat --amt_msat=100000000000 --fiat_backend=coindesk

100000000000 msat = 55668.2283 USD, priced at 2021-04-19 02:00:00 +0200 SAST
```

(This PR kinda clashes with #115 but CoinDesk also has an endpoint for different currency rates so I can update this PR to include that if #115 ends up being accepted 👍 )


fiat/coindesk_api.go
```diff
@@ -0,0 +1,117 @@
...
+// parseCoinDeskData parses http response data from coindesk into USDPrice
+// structs.
+func parseCoinDeskData(data []byte) ([]*USDPrice, error) {
+	var priceEntries coinDeskResponse
+	if err := json.Unmarshal(data, &priceEntries); err != nil {
+		return nil, err
+	}
+
+	// nolint: prealloc
+	var usdRecords []*USDPrice
```
carlaKC, line 61: Why no prealloc?

fiat/prices.go
```diff
@@ -11,11 +11,53 @@ import (
...
+		"fiat prices are enabled")
 )
 
+// PriceAPIBackend is an interface that must be implemented by any backend that
+// is used to fetch fiat price information.
+type PriceAPIBackend interface {
+	GetPrices(ctx context.Context, startTime, endTime time.Time) ([]*USDPrice, error)
+}
+
+type PriceBackend string
```
carlaKC, line 31: nit: godocs for this enum and the coincap/coindesk values below.

I also have a _slight_ preference for uint8 for categories like this, I think it's a bit more standard, but that's a loose opinion loosely held. If you just do ints, you can add a `String` method which returns the names as you have them here.

fiat/coindesk_api.go
```diff
@@ -0,0 +1,117 @@
...
+			Price:     decimal.NewFromFloat(price),
+		})
+	}
+
+	return usdRecords, nil
+}
+
+// GetPrices retrieves price information from coindesks's api for the given
+// time range.
+func (c *coinDeskAPI) GetPrices(ctx context.Context, start,
```
carlaKC, line 80: Time range validation and sort by timestamp is going to be common to all fiat backends, so could we cut the duplication out?

Thinking something like this:
```
type FiatSource struct{
    impl fiatBackend
}

func (f *FiatSource) GetPrices(){
   utils.ValidateTimeRange
   
   records, err := f.impl.rawPriceData()
    
   sort(records)
}
```

Where `fiatBackend` would be coincap/coindesk implementation that just fetches data. You'd need to rename the coincap `GetPrices` and move validation out, but then we can just return a `FiatSource` and we don't even need to export the interface.

frdrpc/exchange_rate.go
```diff
@@ -59,11 +59,24 @@ func granularityFromRPC(g Granularity, disableFiat bool,
 	}
 }
 
+func fiatBackendFromRPC(backend FiatBackend) (fiat.PriceBackend, error) {
+	switch backend {
+	case FiatBackend_UNKNOWN_FIATBACKEND, FiatBackend_COINCAP:
+		return fiat.CoinCapPriceBackend, nil
+	case FiatBackend_COINDESK:
```
carlaKC, line 66: nit: newline between cases makes this a bit more readable

frdrpc/faraday.proto
```diff
@@ -320,6 +320,15 @@ enum Granularity {
     DAY = 8;
 }
 
+/*
+FiatBackend is the API endpoint to be used for any fiat related queries.
+*/
+enum FiatBackend {
+    UNKNOWN_FIATBACKEND = 0;
+    COINCAP = 1;
```
carlaKC, line 328: nit: comments here would be nice, perhaps just list the url that we get this info for? When we generate docs from our protos, the comments on these fields are what end users see in our API docs, so it's nice to be as verbose as possible here. 

Because we don't have docs besides API 🙈

cmd/frcli/utils.go
```diff
@@ -274,3 +275,18 @@ func parseChannelPoint(ctx *cli.Context) (*wire.OutPoint, error) {
 
 	return channelPoint, nil
 }
+
+// parseFiatBackend parses the user chosen fiat backend into a FiatBackend type.
+func parseFiatBackend(fiatBackend string) (frdrpc.FiatBackend, error) {
```
carlaKC, line 280: Ah I see why you went with strings now! Still got some pref for an enum with a `String` method, then this could be 

```
switch fiatBackend{
   case fiat.CoinCapPriceBackend.String(), "":
        return frdrpc.FiatBackend_COINCAP, nil

   case CoinDeskPriceBackend.String():
         return frdrpc.FiatBackend_COINDESK, nil
  
   default: 
       err unknown backend
}
```

fiat/prices.go
```diff
@@ -31,7 +31,7 @@ type PriceAPIBackend interface {
 type PriceBackend string
 
 const (
-	UnknownPriceBackend PriceBackend = ""
+	UnknownPriceBackend  PriceBackend = ""
```
carlaKC, line 34: nit: format in previous commit?

carlaKC (commented):
Awesome job expanding a pretty convoluted part of the codebase! Just a few comments from me re how we abstract our fiat backend implementation.

ellemouton:
Updated 👍

fiat/prices.go
```diff
@@ -6,16 +6,115 @@ import (
...
+var priceBackendNames = map[PriceBackend]string{
+	UnknownPriceBackend:  "",
+	CoinCapPriceBackend:  "coincap",
+	CoinDeskPriceBackend: "coindesk",
+}
+
+func (p PriceBackend) ToString() string {
+	return priceBackendNames[p]
+}
+
```
ellemouton, line 93: hmm so #117 adds another one of these maps. Could get messy if more things are added. How about something like this rather:

```
type backendsConfig struct {
	name string
	proxySupport bool
}

var priceBackendsConfig = map[PriceBackend]backendsConfig {
	UnknownPriceBackend: {},
	CoinCapPriceBackend: {
		name: "coincap",
		proxySupport: false,
	},
	CoinDeskPriceBackend: {
		name: "coindesk",
		proxySupport: true,
	},
}
```

fiat/prices.go
```diff
@@ -6,16 +6,105 @@ import (
...
+	return historicalRecords, nil
+}
+
+// PriceBackend is an enum that indicates which backend we are using for fiat
+// information.
+type PriceBackend int
+
+const (
+	UnknownPriceBackend PriceBackend = iota
```
carlaKC, line 75: nit: either unexport or add godoc?

fiat/prices.go
```diff
@@ -6,16 +6,105 @@ import (
...
+
+	// CoinCapPriceBackend uses CoinCap's API for fiat price data.
+	CoinCapPriceBackend
+)
+
+var priceBackendNames = map[PriceBackend]string{
+	UnknownPriceBackend: "",
```
carlaKC: Probably better to make this "Unknown", otherwise once day we'll forget that it's empty and know why we're logging some random empty string.

fiat/prices.go
```diff
@@ -6,16 +6,105 @@ import (
...
+	CoinCapPriceBackend: "coincap",
+}
+
+func (p PriceBackend) ToString() string {
```
carlaKC, line 90: `String` so that it satisfies `Stringer` interface?

fiat/coindesk_api.go
```diff
@@ -0,0 +1,96 @@
...
+	end time.Time) ([]*USDPrice, error) {
+
+	query := func() ([]byte, error) {
+		return queryCoinDesk(start, end)
+	}
+
+	// CoinDesk uses a granularity of 1 day and does not include the current
+	// day's price information. So subtract 1 period from the start date so
+	// that at least one day's price data is always included.
+	start = start.Add(time.Hour * -24)
```
carlaKC, line 86: nice

frdrpc/faraday.proto
carlaKC: Great docs!

frdrpc/rpcserver.go
```diff
@@ -433,12 +433,12 @@ func (s *RPCServer) ExchangeRate(ctx context.Context,
...
 	log.Debugf("[FiatEstimate]: %v requests", len(req.Timestamps))
 
-	timestamps, granularity, err := parseExchangeRateRequest(req)
+	timestamps, fiatBackend, granularity, err := parseExchangeRateRequest(req)
 	if err != nil {
 		return nil, err
 	}
 
-	prices, err := fiat.GetPrices(ctx, timestamps, *granularity)
+	prices, err := fiat.GetPrices(ctx, timestamps, fiatBackend, *granularity)
```
carlaKC, line 441: These look a bit long, are they wrapped at 80?

carlaKC (commented):
Great commit structure! Makes such a difference in reveiw :)

This is shaping up really well, just some nits, and the imports GH action is failing (experimenting with a check to group imports nicely, there may be some bugs in it so lmk if it seems unreasonable).

ellemouton:
Thanks for the review @carlaKC  😊  updated 👍

fiat/prices.go
```diff
@@ -8,14 +8,107 @@ import (
...
+			historicalRecords[j].Timestamp,
+		)
+	})
+
+	return historicalRecords, nil
+}
+
+// PriceBackend is an enum that indicates which backend we are using for fiat
+// information.
+type PriceBackend int
```
carlaKC, line 73: nit `uint8`?

fiat/prices.go
```diff
@@ -8,14 +8,107 @@ import (
...
+	// CoinCapPriceBackend uses CoinCap's API for fiat price data.
+	CoinCapPriceBackend
 )
 
+var priceBackendNames = map[PriceBackend]string{
+	UnknownPriceBackend: "unknown",
+	CoinCapPriceBackend: "coincap",
+}
+
+func (p PriceBackend) String() string {
```
carlaKC, line 94: nit: godoc

fiat/prices.go
```diff
@@ -8,14 +8,107 @@ import (
...
+	// UnknownPriceBackend is used to indicate that no specific backend
+	// was specified for fiat price data and that the defaults should
+	// instead be used.
+	UnknownPriceBackend PriceBackend = iota
+
+	// CoinCapPriceBackend uses CoinCap's API for fiat price data.
+	CoinCapPriceBackend
 )
 
+var priceBackendNames = map[PriceBackend]string{
```
carlaKC, line 88: ok nice, following the logic for a map here given the context of the follow up PR 👌

carlaKC (commented):
Great set of well structured change! 

Just going to do some testing on my end and then will hit the green button.

carlaKC (approved):
goods as advertised! fiat & audit endpoints flawless with both APIs, and coincap default is backwards compatible 💰 

great improvement to code structure, really good work on this one!

PR #117, opened 2021-04-30
[WIP] multi: add socks proxy flag 

ellemouton:
This PR addresses [this comment suggestion](https://github.com/lightninglabs/faraday/issues/37#issuecomment-671330935) to add a `--proxy` option to proxy http requests for price data over Tor. 

This depends on #116 and so the first 2 commits should be ignored.

It seems to me from testing that I have done than CoinCap does not allow requests over tor, but CoinDesk does. So in this PR the default backend is CoinCap if no proxy is set and is CoinDesk if a proxy is set. If both proxy and backend are specified then a validation check is done to check if the specified backend supports proxy requests.


fiat/prices.go
```diff
@@ -91,17 +93,56 @@ var priceBackendNames = map[PriceBackend]string{
 	CoinDeskPriceBackend: "coindesk",
 }
 
+var priceBackendProxySupport = map[PriceBackend]bool{
```
ellemouton, line 96: thoughts on doing a config map instead of the 2 separate `priceBackendNames` and `priceBackendProxySupport` maps? So something like this:

```
type backendsConfig struct {
	name string
	proxySupport bool
}

var priceBackendsConfig = map[PriceBackend]backendsConfig {
	UnknownPriceBackend: {
		name: "unknown",
		proxySupport: false,
        },
	CoinCapPriceBackend: {
		name: "coincap",
		proxySupport: false,
	},
	CoinDeskPriceBackend: {
		name: "coindesk",
		proxySupport: true,
	},
}
```

carlaKC (commented):
Nice, goods are as advertised! 

Just a unit test then this is gg.

ellemouton:
!lightninglabs-deploy mute

ellemouton:
closing for now - anyone else is welcome to take this over

PR #121, opened 2021-05-20
multi: fix typo and regenerate protos

ellemouton:
This PR consists of 2 unrelated commits:

- One commit to fix a function name typo
- One commit to regenerate the protos (the protobuf compiler version was updated in a previous commit)


carlaKC (approved):
Good catch, LGTM!

PR #122, opened 2021-05-21
multi: custom fiat prices

### LINKED ISSUE #113: Allow for custom price API
If the price API interface is simple enough, we'd like to implement it and use our exchange's data to price the data in the audits.

1. Should be able to select currency other than USD (ie. JPY)
2. Should be able to point the price endpoint to our own API (internal, that supplies historical JPY price data)

ellemouton:
This PR allows a user to specify their own fiat price points in a CSV file in the following format:

Unix Timestamp [seconds] | Price of 1 BTC in chosen currency
------------ | ---------------
1621246440 | 10.1
1621249080 | 123.45
... | ...

Then you can do the following:

```
$ frcli audit --enable_fiat --fiat_backend=custom --prices_csv_path=/path/to/pricedata.csv --custom_price_currency=ZAR --start_time=1621246480 --end_time=1621249110
```

and 

```
$ frcli fiat --amt_msat=100000000000  --fiat_backend=custom --prices_csv_path=/path/to/pricedata.csv --custom_price_currency=ZAR
```

Fixes #113


fiat/customprices.go
```diff
@@ -0,0 +1,18 @@
...
+type customPrices struct {
+	entries []*Price
+}
+
+// rawPriceData returns the custom price point entries.
+func (c *customPrices) rawPriceData(_ context.Context, _,
+	_ time.Time) ([]*Price, error) {
+
+	return c.entries, nil
+}
```
ellemouton, line 18: Could filter the data given the times here but seems unnecessary? If anything, maybe they should be filtered in frcli so that only the necessary data is sent across?

cmd/frcli/csv.go
```diff
@@ -10,10 +14,10 @@ import (
 // CSVHeaders returns the headers used for harmony csv records.
 var CSVHeaders = "Timestamp,OnChain,Type,Category,Amount(Msat),Amount(Fiat),TxID,Reference,BTCPrice,BTCTimestamp,FiatCurrency,Note"
 
-// csv returns a csv string of the values contained in a rpc entry. For ease
+// writeToCSV returns a csv string of the values contained in a rpc entry. For ease
 // of use, the credit field is used to set a negative sign (-) on the amount
 // of an entry when it decreases our balance (credit=false).
-func csv(e *frdrpc.ReportEntry) string {
+func writeToCSV(e *frdrpc.ReportEntry) string {
```
ellemouton, line 20: is this rename fine? 😅  only did it to that i could import the csv package without renaming the import. But can defs just rename the import if that would be better

accounting/conversions.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/lightninglabs/faraday/utils"
 )
 
-// usdPrice is a function which gets the USD price of bitcoin at a given time.
-type usdPrice func(timestamp time.Time) (*fiat.USDPrice, error)
+// fiatPrice is a function which gets the fiat price of bitcoin at a given time.
+type fiatPrice func(timestamp time.Time) (*fiat.Price, error)
```
carlaKC, line 14: nit: some of these rename refactors could be split out into their own commits, makes things easy to catch.

They certainly should have been generically named in the first place, but here we are, sigh.

fiat/coincap_api_test.go
```diff
@@ -189,14 +189,16 @@ func TestParseCoinCapData(t *testing.T) {
...
+	expectedPrices := []*Price{
 		{
 			Price:     price1,
 			Timestamp: time1,
+			Currency:  "USD",
 		},
 		{
 			Price:     price2,
 			Timestamp: time2,
+			Currency:  "USD",
```
carlaKC, line 201: nit: could be const

fiat/fiat.go
```diff
@@ -23,21 +23,26 @@ var (
 	errRetriesFailed = errors.New("could not get data within max retries")
 )
 
-// USDPrice represents the Bitcoin price in USD at a certain time.
-type USDPrice struct {
+// Price represents the Bitcoin price in the given Currency at a certain time.
```
carlaKC, line 26: s/Currency/currency

frdrpc/faraday.proto
```diff
@@ -360,11 +360,15 @@ message ExchangeRateResponse {
...
-    // The price of 1 BTC, expressed in USD.
+    // The price of 1 BTC, expressed in the currency denoted in the currency
+    // field.
     string price = 1;
 
     // The timestamp for this price price provided.
     uint64 price_timestamp = 2;
+
+    // The currency that the price is denoted in.
+    string currency = 3;
```
carlaKC, line 377: Wondering whether this should be an enum rather than a string? Perhaps not worth the effort, WDYT?

cmd/frcli/csv.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // CSVHeaders returns the headers used for harmony csv records.
-var CSVHeaders = "Timestamp,OnChain,Type,Category,Amount(Msat),Amount(USD),TxID,Reference,BTCPrice,BTCTimestamp,Note"
+var CSVHeaders = "Timestamp,OnChain,Type,Category,Amount(Msat),Amount(Fiat),TxID,Reference,BTCPrice,BTCTimestamp,FiatCurrency,Note"
```
carlaKC, line 15: I think that we can be pretty certain that the report will always be generated with a single fiat currency, so we probably don't need to repeat it in a `FiatCurrency` column for each row. We could just insert our currency string in `Amount(Fiat)` where we [use this value](https://github.com/lightninglabs/faraday/blob/master/cmd/frcli/node_audit.go#L197).

fiat/customprices.go
carlaKC: Yeah that sounds like a good idea!

fiat/prices.go
```diff
@@ -37,6 +37,22 @@ type PriceSource struct {
 	impl fiatBackend
 }
 
+// PriceSourceConfig is a struct holding various config options used for
+// initialising a new PriceSource.
+type PriceSourceConfig struct {
```
carlaKC, line 45: Add a validation method now that we're pulling this out into a separate function? 

eg, coincap has to have granularity, custom backend must have some price points etc

cmd/frcli/csv.go
carlaKC: yeah sure, not the finest choice of name if the first place! would separate it out into another commit, because I'm fussy like that, but def not required.

cmd/frcli/csv.go
```diff
@@ -27,3 +31,55 @@ func csv(e *frdrpc.ReportEntry) string {
...
+	// to contain column headers.
+	csvLines = csvLines[1:]
+
+	prices := make([]*frdrpc.BitcoinPrice, len(csvLines))
+
+	for i, line := range csvLines {
+		if len(line) < 2 {
+			return nil, errors.New("incorrect csv format. " +
+				"At least two columns items are expected per " +
+				"row")
```
carlaKC, line 69: Can make this stricter to require exactly 2 columns? Don't want people give weird junk input with extra columns.

frdrpc/node_audit.go
```diff
@@ -53,9 +55,15 @@ func parseNodeAuditRequest(ctx context.Context, cfg *Config,
 		return nil, nil, err
 	}
 
+	pricePoints, err := pricePointsFromRPC(req.CustomPrices)
```
carlaKC, line 58: Probably also want to only accept custom prices if `CUSTOM` was set?

carlaKC (commented):
Looking great! 

Mostly comments re how we validate this new input, could use slightly stricter checks so it's very clear what we're requiring. 
Sorry it took me so long! Also could move some of the renames into their own commits, but that is a hassle so non-blocking.

frdrpc/faraday.proto
ellemouton: would that not mean that we need to add an enum entry for each possible currency code option that a user could want to use? If it is a string then they can use any arbitrary currency. right?

carlaKC: Oh yeah ofc, ignore!

frdrpc/node_audit.go
ellemouton: do you mean like error out if they provide prices but dont use custom option?

ellemouton:
Updated and split out some commits. I still gotta add the price point time stamp filtering on the frcli side. Will do that soon 👍

ellemouton:
cool, the last commit does the timestamp filtering in frcli

cmd/frcli/fiat_estimate.go
```diff
@@ -67,24 +67,30 @@ func queryFiatEstimate(ctx *cli.Context) error {
...
 	// Set start and end times from user specified values, defaulting
 	// to zero if they are not set.
 	req := &frdrpc.ExchangeRateRequest{
 		Timestamps:   []uint64{uint64(ts)},
 		FiatBackend:  fiatBackend,
-		CustomPrices: customPrices,
+		CustomPrices: filteredPrices,
 	}
 
 	rpcCtx := context.Background()
```
ellemouton: ok so technically now with this filtering (in the case of Custom price source backend) only 1 timestamp will be sent to faraday and will be returned as is .... so for fiat_estimate (given custom option) faraday doesnt do anything... so the call to faraday could just be skipped here and all processing can be done in frcli (even in the case of multiple timestamps). Then the `CustomPrices` field in the `ExchangeRateRequest` can also be removed i think. Let me know what you think . The upside, i think, is an unnecessary network call and the downside is some code bloat and possible duplicate logic (in the case of multiple timestamps)

fiat/coincap_api.go
```diff
@@ -15,6 +15,10 @@ import (
 const (
 	// coinCapHistoryAPI is the endpoint we hit for historical price data.
 	coinCapHistoryAPI = "https://api.coincap.io/v2/assets/bitcoin/history"
+
+	// coinCapDefaultCurrency is the currency that the price data returned by the
```
carlaKC, line 19: is this wrapped at 80? Looks a leetle off to me.

cmd/frcli/node_audit.go
```diff
@@ -194,6 +194,13 @@ func queryOnChainReport(ctx *cli.Context) error {
 		}
 	}()
 
+	if len(report.Reports) > 0 {
+		CSVHeaders = strings.Replace(
```
carlaKC, line 238: Instead of `strings.Replace` we could have: 

```
var CSVHeaders = "Timestamp,OnChain,Type,Category,Amount(Msat),Amount(%v),TxID,Reference,BTCPrice,BTCTimestamp,Note"

headers = fmt.Sprintf(CSVHeaders, report.Reports[0].BtcPrice.Currency)
```

Just a little shorter, and gives us options to add more variable fields going forward.

accounting/config.go
```diff
@@ -131,12 +127,11 @@ func NewOnChainConfig(ctx context.Context, lnd lndclient.LndServices, startTime,
...
 		CommonConfig: CommonConfig{
-			StartTime:   startTime,
-			EndTime:     endTime,
-			DisableFiat: disableFiat,
-			FiatBackend: fiatBackend,
-			Granularity: granularity,
-			Categories:  categories,
+			StartTime:      startTime,
+			EndTime:        endTime,
+			DisableFiat:    disableFiat,
```
carlaKC, line 132: Do you think it makes sense for a nil price source cfg to imply disable fiat? I'm on the fence, because nilness checks can be a bit ambiguous.

fiat/prices.go
```diff
@@ -37,6 +43,44 @@ type PriceSource struct {
...
+	Granularity *Granularity
+
+	// PricePoints is a set of price points that is used for fiat related
+	// queries if the PriceBackend being used is the CustomPriceBackend.
+	PricePoints []*Price
+}
+
+// validatePriceSourceConfig checks that the PriceSourceConfig fields are valid
+// given the chosen price backend.
+func validatePriceSourceConfig(cfg *PriceSourceConfig) error {
```
carlaKC, line 64: could also be `func (p *PriceSourceConfig) validate()error` and then we only validate if non-nil.

frdrpc/node_audit.go
carlaKC: yep! rather be really strict with what we're accepting so there's no confusion :)

cmd/frcli/fiat_estimate.go
carlaKC: I think that it's ok to leave the call in. There's a use case for throwing a ton of custom price data at faraday and wanting it to pick the right timestamp + do duplication. I'd err on the side of not adding custom logic to save ourselves one rpc call.

carlaKC (commented):
A few last nits from me, looking spectacular. 

I like having the `frcli` side filtering, saves us from shuffling a ton of unnecessary data back and forth. 

The last thing that I think we need is to add some `rpcserver` side validation of the custom price data. The way that our price logic works is that we _always_ need a timestamp that is before our start date to be able to get prices. This should live in the server, so that cli and rpc clients are validated. Since prices can be provided at varying granularity (eg, yearly, monthly, daily), it's difficult to know what other checks we should enforce on that data, but at least <= start date should be required.

accounting/config.go
ellemouton: hmmm i think not cause priceCfg will never be nil since it will be created with default values if no specific options are provided. Thoughts?

ellemouton: i guess, could also force priceCfg to be nil earlier on if fiat is disabled?

carlaKC: > i guess, could also force priceCfg to be nil earlier on if fiat is disabled?

yeah that was what I was thinking, `if disableFiat, priceCfg=nil`, but I think it'll just be cleaner to keep the bool + ignore if we're disabled. Just a errant thought :)

fiat/prices.go
```diff
@@ -37,6 +43,40 @@ type PriceSource struct {
...
+	// PricePoints is a set of price points that is used for fiat related
+	// queries if the PriceBackend being used is the CustomPriceBackend.
+	PricePoints []*Price
+}
+
+// validatePriceSourceConfig checks that the PriceSourceConfig fields are valid
+// given the chosen price backend.
+func (cfg *PriceSourceConfig) validatePriceSourceConfig() error {
+	switch cfg.Backend {
+	case UnknownPriceBackend, CoinCapPriceBackend:
```
carlaKC, line 66: Why do we require granularity set for `UnknownPriceBackend`? Wouldn't we only expect it for coincap?

carlaKC (approved):
Awesome to wrap this up! 

Just one Q left, probably because I've forgotten context here :')

PR #123, opened 2021-05-31
Macaroon and log fixes

### LINKED ISSUE #84: Don't require all macaroons to use Faraday

guggero:
Fixes #84.
Allows faraday to only use the readonly.macaroon (or a custom baked one)
to connect to lnd instead of needing to copy all macaroons (including
the subserver ones).

We also fix an issue that popped up during testing of the latest LiT release where the faraday log level would overwrite any other log level set beforehand.

#### Pull Request Checklist
- [X] Update `release_notes.md` if your PR contains major features, breaking changes or bugfixes


carlaKC (approved):
LGTM!

junderw:
Just out of curiosity, what would be the least permissions (ie. which specific endpoints does it use from lnd) for this?

guggero:
There is no definitive list yet. From a quick glance at where the `lndclient.LndServices` instance is used, it looks like `GetInfo, ListChannels, ClosedChannels, PendingChannels, ListTransactions, ListSweeps, ListForwards, ListInvoices, ListPayments and DecodePayReq` are used. So maybe you could bake a macaroon with `onchain:read, offchain:read, info:read` and you'd be fine.

We're planning on adding something like "macaroon recipes" to Faraday, Loop and Pool so you can bake exactly the macaroon you need. But for that to be as dynamic as possible, we first need to figure out a way to determine this list at build time somehow.

PR #129, opened 2021-08-30
fiat: add coingecko source

carlaKC:
#### Pull Request Checklist
- [x] Update `release_notes.md` if your PR contains major features, breaking changes or bugfixes


fiat/prices.go
```diff
@@ -63,9 +66,17 @@ type PriceSourceConfig struct {
 // given the chosen price backend.
 func (cfg *PriceSourceConfig) validatePriceSourceConfig() error {
 	switch cfg.Backend {
-	case UnknownPriceBackend, CoinCapPriceBackend:
+	// If an unknown price backend is set, we don't expect granularity to
+	// be set. The caller should provide a backend if they wish to use
+	// a specific granularity.
+	case UnknownPriceBackend:
```
ellemouton, line 72: or coinDesk?

ellemouton (approved):
tACK! Just one comment. Also, why not use coingecko as the default since it is more granular than coindesk?

carlaKC:
> tACK! Just one comment. Also, why not use coingecko as the default since it is more granular than coindesk?
> 
> EDIT: oh, i see it's probably cause they only provide data for the last 90 days?

Yeah, pretty inefficient way to fetch data so I thought it made sense to just go with the daily price.

fiat/prices.go
carlaKC: Nice catch!

Added a separate case for coindesk to allow setting daliy granularity, since that's what it provides, but failing otherwise + updated unit tests.

carlaKC:
Thx for quick review on this, really appreciated! Just pushed the suggested changes, diff [here](https://github.com/lightninglabs/faraday/compare/cf5462fd8e3d4f780fca4d671478f6dda1ee160a..0979587eeb3d39c635676573fd5dc0cac66eb124).

PR #140, opened 2022-01-14
multi: use lndclient MacaroonService

ellemouton:
Since the code for creating and using a macaroon service is the same for
multiple projects (pool, loop, faraday & litd), the code has been
unified in lndclient. So this commit removes the macaroon service code
and instead uses the lndclient code.

Depends on https://github.com/lightninglabs/lndclient/pull/86


carlaKC (approved):
utACK

guggero:
Tested this and seems to work well with upgrading/changing the password.
Though in Faraday this has the effect that now the `readonly.macaroon` isn't enough anymore to connect to `lnd`, because deriving the shared key requires a `write` permission.

This also causes the itest to fail:

```
[faraday-err] Error starting faraday: error starting macaroon service: unable to derive a shared secret with LND: rpc error: code = Unknown desc = permission denied
```

Not sure if this is a problem, @carlaKC ? If it is, we could say that since Faraday does read only stuff only (currently), encrypting the macaroon DB isn't as important as in the other daemons?

carlaKC:
> Though in Faraday this has the effect that now the readonly.macaroon isn't enough anymore to connect to lnd, because deriving the shared key requires a write permission.

I think it's ok to just switch over to needing write permissions for this purpose? If we update the docs accordingly, shouldn't be an issue.

guggero (changes requested):
> I think it's ok to just switch over to needing write permissions for this purpose? If we update the docs accordingly, shouldn't be an issue.

Okay, cool. Just wanted to let you know in case there are complaints later on...

@ellemouton seems like this requires a small change in the itest to no longer use the `readonlym.macaroon`. And we should also update the docs where necessary.

ellemouton:
updated 👍  thanks for catching!

guggero (approved):

guggero:
Argh, another PR that fell off the radar, sorry. Can you rebase please? Then this can get in before https://github.com/lightninglabs/faraday/pull/147.

ellemouton:
rebased :)

junderw:
This was unfortunate...

PR #146, opened 2022-06-14
frdrpcserver: move RequiredPermissions to dedicated dir

ellemouton:
This commit moves the RequiredPermissions map to its own directory so
that projects importing the permissions list dont need to import all the
dependencies of the frdrpcserver package.


guggero (approved):
Linter complains about the formatting, otherwise LGTM :tada:

PR #148, opened 2022-06-24
frdrpcserver: also create macaroon service in subserver mode

guggero:
This fixes a small oversight in #140 where we forgot to also _create_
the macaroon service in the StartAsSubserver() method.

#### Pull Request Checklist


ellemouton (approved):
ah yes, good catch. Sorry about the oversight! LGTM!

PR #155, opened 2023-01-10
fiat: add support fo multiple granularities to coingecko

positiveblue:
The coingecko API automatically changes its granularity depending on how old is the pricing data you are asking for. For 90 days or less, it returns hourly entries while for any date 91 days or older it returns daily granularity.

To obtain the most exact results we should always make any reports in that 90 days window. That's why we currently do not support ranges that include dates older than 90 days.

In some case it may make sense to use daily data. This commit extends the current API to adapt the range and ensure that we get always the best possible data.

#### Pull Request Checklist

### DIFF
--- a/fiat/coingecko.go
+++ b/fiat/coingecko.go
@@ -18,27 +18,8 @@ const (
 	defaultCoinGeckoCurrency = "USD"
 )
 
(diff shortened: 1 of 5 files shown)


guggero (approved):
Very nice, LGTM :tada:

Issue #163, opened 2023-04-19
Bug: Can't send custom_prices over REST protocol


junderw:
**Expected behavior**

The ability to send a request with a CUSTOM backend and custom_prices via REST.
[Docs here say REST position unknown](https://lightning.engineering/api-docs/api/faraday/faraday-server/node-audit/index.html#frdrpcnodeauditrequest)

**Actual behavior**

I can't figure out a proper way to send the request.

**To reproduce**

For example: (change values accordingly)

```bash
curl \
    -k \
    -g \
    -H 'Grpc-Metadata-macaroon:'$(cat ~/.faraday/regtest/faraday.macaroon | xxd -ps -c 99999999) \
    https://localhost:8466/v1/faraday/nodeaudit?start_time=1674089890\&fiat_backend=CUSTOM\
\&custom_prices[0][price]=4000000\
\&custom_prices[0][price_timestamp]=1674089870\
\&custom_prices[0][currency]=JPY\
\&custom_prices[1][price]=4100000\
\&custom_prices[1][price_timestamp]=1674089930\
\&custom_prices[1][currency]=JPY\
 && echo ""
```

Changing custom_prices to something like `custom_prices=glorp` it gives me a parse error for BtcPrices... so it's expecting it via the query parameters.

The solution might be to somehow allow this data to be included in the body of the request as JSON or something...

Even if we somehow got this working, 108 characters per entry will blow up and hit some URL query limits eventually...

guggero:
Thanks for the report. This is a problem indeed. We shouldn't have used `GET` for all calls, as the array type is not supported by the library that does the REST to gRPC mapping. I added additional mappings for `POST` for most of the calls, see https://github.com/lightninglabs/faraday/pull/164. Can you confirm that works for you?

ellemouton:
@junderw , in case it helps, the following request works for me with @guggero 's fix:

```
curl \
    -vk \
    -H 'Grpc-Metadata-macaroon:'$(cat "/path/to/faraday.macaroon" | xxd -ps -c 99999999) \
    https://localhost:9001/v1/faraday/nodeaudit -d '{"start_time": "1681893194","fiat_backend":"CUSTOM", "customPrices": [{"price":"11", "price_timestamp":"1621246440", "currency":"ZAR"}, {"price":"123", "price_timestamp":"1621249080", "currency":"ZAR"}, {"price":"111", "price_timestamp":"1681728012", "currency":"ZAR"}]}' | jq

```

[...]

PR #164, opened 2023-04-19
frdrpc: add POST bindings for more complex calls

guggero:
Fixes https://github.com/lightninglabs/faraday/issues/163.

Using GET mappings for REST means we have to place every parameter in the URL query. This is not very convenient if there are many parameters. But even worse, some types like arrays/slices aren't supported in query parameters and map types are very awkward to use.
So we add POST mappings for the calls that have many or more complex request parameters.

#### Pull Request Checklist


bhandras (approved):
LGTM 📮

ellemouton (approved):
tACK 🔥

Issue #177, opened 2023-12-07
ResourceExhausted error while running the Node Audit


remnantdev:
We run into a ResourceExhausted error while running the NodeAudit report.

```
Error received from peer ipv4:10.9.82.14:8465","file":"/var/local/git/grpc/src/core/lib/surface/call.cc","file_line":1075,"grpc_message":"grpc: received message larger than max (214449486 vs. 209715200)","grpc_status":8}
```

- The error does not seem to be dependant on the requested time range.
- We run the report with `DisableFiat:true` and `UnknownGranularity`.
- This error occurs on only one of our nodes presumably the one with the most onchain transactions.
  - We do use our LND node to send and receive onchain transactions
  - Looking through the [report code](https://github.com/lightninglabs/faraday/blob/71bb06f6f7776d21ff4d600fcb41f10cdf33c300/accounting/config.go#L123), it seems that all transactions are requested from the LND Server in a single RPC call which could be over the `maxMsgRecvSize` of  200MB and would explain it occurs no matter what the requested time range.

**Expected behavior**

Node Audit should produce the report.

**Actual behavior**

Node Audit should produce the report 

**To reproduce**

This error now occurs systematically on one node when we run Node Audit

**System information**

- Connecting through gRPC
- Version: Faraday v0.2.11-alpha

guggero:
Hey, thanks for the report. You are correct, the receive size is too small. We should really start paginating calls to `ListTransactions`, but in the meantime, can you please try with https://github.com/lightninglabs/faraday/pull/178?

guggero:
I pushed a new commit that adds the `--lnd.requesttimeout` config options that takes a duration. So you can bump it to `--lnd.requesttimeout=2m` or whatever you feel is appropriate (note the time unit suffix, without a unit I believe this is in nano-seconds).

PR #178, opened 2023-12-07
accounting: increase client gRPC max receive size, bump lndclient to fix large message issue

guggero:
~Depends on https://github.com/lightninglabs/lndclient/pull/164.~

Fixes https://github.com/lightninglabs/faraday/issues/177, at least temporarily. We should start paginating calls to `ListTransactions`.

#### Pull Request Checklist


mrfelton:
We're using this successfully to enable us to work with large nodes that have a lot of data. Good to merge imo, though definitely we pagination is needed as the memory requirements are ever increasing with the current implementation that fetches every single onchain transaction.

guggero:
Updated to use the tagged version of `lndclient`.

bhandras (approved):

Issue #137, opened 2021-12-07
Report fees for onchain transactions that split UTXOs


giancarloGiuffra:
Onchain transactions that only pay to addresses controlled by the node are not reported. From the point of view of accounting this seems correct. However, the paid onchain fees are not reported either. These fees affect the onchain balance of the node and thus should be included in the audit report.

Zero-amount onchain transactions are explicitly excluded here:
https://github.com/lightninglabs/faraday/blob/12b66ac753c92d0c237ad2523f0ec93c82898f92/accounting/entries.go#L245-L261

I agree that any sweep transaction of the remote counterparty should not be included since it was the remote counterparty that paid the fees. However, that condition is also excluding the transactions I described above.

mrfelton:
It would be great for faraday to be able to report on these pay to self sweeps. Without this, the reports generated by faraday is missing fees paid when doing UTXO consolidation or splitting, and so calculations based of those reports appear to show higher node balances than in reality and miss miss the full picture. Over time the discrepancies widen as the differences between what faraday reports and the actually reality of what happened on the build up.

PR #180, opened 2024-01-27
Record fees for self initiated sweeps to self

mrfelton:
Create onchain fee entry for self initiated sweeps that pay to self. Ensures that fees relating to UTXO consolidation or splitting are included in the audit reports.

Fixes https://github.com/lightninglabs/faraday/issues/137

Looking for a concept ack.

Open questions:

1. What would be the most reliable way to identify these types of pay to self zero amount sweep transactions? The approach I'm using is to inspect all the inputs and outputs and verify that they belong to the us.
2. What entries should we have in the report for these tyoes of transactions? Approach here is is to just create a single onchain fee entry. But maybe we should also have an onchain payment and an onchain receive to go with it?


accounting/entries.go
```diff
@@ -231,15 +231,60 @@ func sweepEntries(tx lndclient.Transaction, u entryUtils) ([]*HarmonyEntry, erro
...
+			return false
+		}
+	}
+
+	// If all inputs and outputs belong to our wallet, it's utxo management.
+	return true
+}
+
+// createOnchainFeeEntry creates a fee entry for an on chain transaction.
+func createOnchainFeeEntry(tx lndclient.Transaction, category string, note string, u entryUtils) (*HarmonyEntry, error) {
```
guggero, line 255: nit: line length (we're trying to adhere to the same code formatting rules as [in `lnd`](https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md#80-character-line-length), including the 80 character line length limit.

accounting/entries.go
```diff
@@ -231,15 +231,60 @@ func sweepEntries(tx lndclient.Transaction, u entryUtils) ([]*HarmonyEntry, erro
 	return []*HarmonyEntry{txEntry, feeEntry}, nil
 }
 
+// isUtxoManagementTx checks whether a transaction is restructuring our utxos.
+func isUtxoManagementTx(txn lndclient.Transaction) bool {
```
guggero, line 235: I think this probably works well for the intended purpose. Though I wonder, how did such a transaction show up in `faraday` before? Was it just ignored or booked under another category? I'm not super familiar with all the functionality of `faraday` unfortunately...

accounting/entries_test.go
```diff
@@ -569,8 +593,14 @@ func TestOnChainEntry(t *testing.T) {
 		// Whether the transaction has a fee attached.
 		hasFee bool
 
+		// Whether the transaction is a sweep
```
guggero, line 596: nit: missing full stop.

guggero (commented):
Thanks for the PR. I think this gets the job done. Can't really think of an other or easier way to detect a manual consolidation TX.

Just to complete my understanding, how did such a transaction show up in `faraday` before?
Maybe it would be worth adding the scenario to the integration tests to make sure the result is what we expect.

Also, can you please squash your commits? Thanks.

mrfelton:
> Just to complete my understanding, how did such a transaction show up in faraday before?

It didn't. It was totally omitted.

> Maybe it would be worth adding the scenario to the integration tests to make sure the result is what we expect.

I have added tests that prove that these types of transactions are now included in the report.

accounting/entries.go
mrfelton: It was totally omitted, resulting in under-reporting on fees by faraday.

accounting/entries_test.go
mrfelton: Added

accounting/entries.go
mrfelton: Updated.

mrfelton:
> Also, can you please squash your commits? Thanks.

Comments addressed and commits squashed.

accounting/entries.go
```diff
@@ -231,15 +231,65 @@ func sweepEntries(tx lndclient.Transaction, u entryUtils) ([]*HarmonyEntry, erro
...
+			return false
+		}
+	}
+
+	// If all inputs and outputs belong to our wallet, it's utxo management.
+	return true
+}
+
+// createOnchainFeeEntry creates a fee entry for an on chain transaction.
+func createOnchainFeeEntry(
```
guggero, line 255: Can you please take a look at the formatting guideline doc I shared? Or the surrounding code? This isn't how we want our function definitions to be formatted.
Alternatively I can also push a fix commit to your branch if you don't want to deal with it.

mrfelton: I missed the link in your prior comment. I've pushed an update with that style fix. Shame that `make fmt` doesn't handle that properly.

guggero: Thanks!

> Shame that `make fmt` doesn't handle that properly.

Yeah, definitely. One of the biggest downsides of Go is that there is no configurable formatter such as `prettier`...

mrfelton:
Thanks @bhandras I've addressed the style nits and force pushed an updated squeezed commit

Issue #181, opened 2024-02-09
Include invoice description from outgoing payments in audit notes


mrfelton:
The `audit` report includes the invoice `description` as a `memo` in the `note` field for incoming payment to an invoice that we issued. However, it does not include the invoice `description` in entries that relate to outgoing payments that we make to an invoice that was provided to us.

Invoice descriptions can provide important context on payments that are useful for reconciliation purposes.

We should include the descriptions for outgoing payments, as we do for incoming ones.

PR #182, opened 2024-02-09
Include outgoing payment description in audit memo

mrfelton:
Include invoice memo in audit nodes for outgoing payments that pay to an invoice. This allows for more options to correlate money movements during reconciliation.

Context: Improves ability to reconcile payments related to peerswaps. See https://github.com/ElementsProject/peerswap/issues/254 for additional context.

This pull request involves changes to the `accounting` package, specifically the `entries.go`, `entries_test.go`, `filter.go`, and `filter_test.go` files. The changes primarily involve the addition of a `description` field to the `paymentInfo` struct and the modification of the `paymentNote` function to include the `memo` field. The changes also involve modifications to the `paymentEntry` and `paymentRequestDetails` functions to accommodate the new `description` field.

Fix https://github.com/lightninglabs/faraday/issues/181

Changes to `entries.go` and `entries_test.go`:

* `func paymentReference(sequenceNumber uint64, preimage lntypes.Preimage) string {`: Modified the `paymentNote` function to include a `memo` field and to append the `memo` and `dest` fields to a `notes` slice. The function now returns a string joined by the `notes` slice elements instead of just the `dest` string.
* `func paymentEntry(payment paymentInfo, paidToSelf bool,`: Modified the `paymentNote` function call in the `paymentEntry` function to include the `payment.description` field.
* [`func TestPaymentEntry(t *testing.T) {`](diffhunk://#diff-3162c2e2d764199457c98cf2788a4e46ced2105b67eddb8af995c6cd13e61f8fL761-R759): Modified the `paymentNote` function call in the `TestPaymentEntry` test function to include the `&invoiceMemo` field. [[1]](diffhunk://#diff-3162c2e2d764199457c98cf2788a4e46ced2105b67eddb8af995c6cd13e61f8fL761-R759) [[2]](diffhunk://#diff-3162c2e2d764199457c98cf2788a4e46ced2105b67eddb8af995c6cd13e61f8fL776-R774)

Changes to `filter.go` and `filter_test.go`:

* [`type paymentInfo struct {`](diffhunk://#diff-0b49c6914e8b2658a934a8474133d297cb3ee0e68ce24b7d61d32e9922e518fcL112-R121): Added a `description` field to the `paymentInfo` struct.
* `func preProcessPayments(payments []lndclient.Payment,`: Modified the `preProcessPayments` function to include the `description` field in the `paymentInfo` struct.
* `func paymentRequestDetails(paymentRequest string, decode decodePaymentRequest) (*route.Vertex, *string, error) {`: Renamed the `paymentRequestDestination` function to `paymentRequestDetails` and modified it to return a `description` field in addition to the `destination` field.
* [`func TestPaymentRequestDestination(t *testing.T) {`](diffhunk://#diff-eab7db6872cddeafbec44ce2a4d2f5b6e5af13c8b7b9f7d3708359f076ee840cL440-R473): Renamed the `TestPaymentRequestDestination` test function to `TestPaymentRequestDetails` and modified it to test the `description` field in addition to the `destination` field. [[1]](diffhunk://#diff-eab7db6872cddeafbec44ce2a4d2f5b6e5af13c8b7b9f7d3708359f076ee840cL440-R473) [[2]](diffhunk://#diff-eab7db6872cddeafbec44ce2a4d2f5b6e5af13c8b7b9f7d3708359f076ee840cL479-R490)

### DIFF
--- a/accounting/entries.go
+++ b/accounting/entries.go
@@ -401,11 +401,18 @@ func paymentReference(sequenceNumber uint64, preimage lntypes.Preimage) string {
 
 // paymentNote creates a note for payments from our node.
 // nolint: interfacer
-func paymentNote(dest *route.Vertex) string {
-	if dest == nil {
-		return ""
+func paymentNote(dest *route.Vertex, memo *string) string {
+	var notes []string
+
+	if memo != nil && *memo != "" {
+		notes = append(notes, fmt.Sprintf("memo: %v", *memo))
 	}
-	return dest.String()
+
+	if dest != nil {
+		notes = append(notes, fmt.Sprintf("destination: %v", dest))
+	}
+
+	return strings.Join(notes, "/")
 }
 
 // paymentEntry creates an entry for an off chain payment, including fee entries
@@ -432,7 +439,7 @@ func paymentEntry(payment paymentInfo, paidToSelf bool,
 
 	// Create a note for our payment. Since we have already checked that our
 	// payment is settled, we will not have a nil preimage.
-	note := paymentNote(payment.destination)
+	note := paymentNote(payment.destination, payment.description)
 	ref := paymentReference(payment.SequenceNumber, *payment.Preimage)
 
 	// Payment values are expressed as positive values over rpc, but they
--- a/accounting/filter.go
+++ b/accounting/filter.go
@@ -109,15 +109,16 @@ func filterInvoices(startTime, endTime time.Time,
 	return filtered
 }
 
-// paymentInfo wraps a lndclient payment struct with a destination, if it is
-// available from the information we have available, and its settle time.
-// Since we now allow multi-path payments, a single payment may have multiple
-// htlcs resolved over a period of time. We use the most recent settle time for
-// payment because payments are not considered settled until all the htlcs are
-// resolved.
+// paymentInfo wraps a lndclient payment struct with a destination, and
+// description if available from the information we have available, and its
+// settle time. Since we now allow multi-path payments, a single payment may
+// have multiple htlcs resolved over a period of time. We use the most recent
+// settle time for payment because payments are not considered settled until
+// all the htlcs are resolved.
 type paymentInfo struct {
 	lndclient.Payment
 	destination *route.Vertex
+	description *string
 	settleTime  time.Time
 }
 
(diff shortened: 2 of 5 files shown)


accounting/filter.go
```diff
@@ -135,24 +136,27 @@ func preProcessPayments(payments []lndclient.Payment,
...
-		// all payments, so we do not error if it is not.
-		destination, err := paymentHtlcDestination(payment)
-		if err != nil {
-			destination, err = paymentRequestDestination(
-				payment.PaymentRequest, decode,
-			)
-			if err != nil && err != errNoPaymentRequest {
-				return nil, err
-			}
+		// Attempt to obtain the payment destination and description from our
```
guggero, line 139: nit: we use 8 spaces per tab character, so these lines are slightly too long.

accounting/filter.go
```diff
@@ -135,24 +136,27 @@ func preProcessPayments(payments []lndclient.Payment,
...
+		// did not store HTLCs.
+		destination, description, err := paymentRequestDetails(
+			payment.PaymentRequest, decode,
+		)
+		if err != nil && err != errNoPaymentRequest {
+			return nil, err
+		}
+
+		if destination == nil {
+			destination, _ = paymentHtlcDestination(payment)
```
guggero, line 153: I think this change alters the preference. Before, the data from `paymentHtlcDestination` took preference over `paymentRequestDestination()`. Now it's the other way around.

guggero (commented):
Thanks for the improvement!
One question about order of precedence.

accounting/filter.go
mrfelton: We need to inspect the invoice in order to get the description, so given that we will always be inspecting the invoice, it seems that inspecting the HTLCs only makes sense as a fallback in case we weren't able to get the destination from the payment request.

Is there an expectation that the destination obtained from the HTLC's could be different from the destination in the payment request?

guggero: Actually, I'm not sure. That's why I'm not really comfortable changing the logic here. Would need to dig into the lnd code and docs to find out. But maybe @carlaKC has a quick answer for me?

mrfelton: I've pushed an additional commit which returns the original order of precedence for the destination

https://github.com/lightninglabs/faraday/pull/182/commits/c478467268092ebea1efc38f497ae115015f315f

It does add some additional overhead though since in many cases we'd be inspecting HTLCs to find the destination even though we already know it from inspecting  the payment request.

guggero (approved):
thanks, LGTM :tada:

accounting/entries.go
```diff
@@ -401,11 +401,22 @@ func paymentReference(sequenceNumber uint64, preimage lntypes.Preimage) string {
...
+
+	if memo != nil && *memo != "" {
+		notes = append(notes, fmt.Sprintf("memo: %v", *memo))
+	}
+
+	if dest != nil {
+		notes = append(notes, fmt.Sprintf("destination: %v", dest))
+	}
+
+	if len(notes) == 0 {
```
bhandras, line 415: nit: no need for this check as `strings.Join()` will just return an empty string if `notes` is empty.

accounting/filter.go
bhandras: Could you please squash the two commits?

bhandras (approved):
LGTM, thanks @mrfelton 🎉

accounting/entries.go
mrfelton: Updated in most recent squashed commit.

PR #192, opened 2024-08-15
config+frcli: fix macaroon & tls cert path parsing

### LINKED ISSUE #191: Unable to define --macaroonpath when starting service
I first noticed this behavior with litd, but it also exists in Faraday. Faraday fails to start if you define the `--macaroonpath` argument.

**Expected behavior**

I would expect to be able to define any directory for macaroonpath and have that directory be used to store faraday's macaroons. Per the usage help text:
```
Path to write the macaroon for faraday's RPC and REST services if it doesn't exist. (default: /Users/joshfix/Library/ApplicationSupport/Faraday/mainnet/faraday.macaroon)
```

**Actual behavior**

When specifying the macaroon path, the service fails to start with the error:
```
Error starting faraday: error validating config: faradaydir overwrites macaroonpath, please only set one value
```
This error occurs even when not specifying a `faradaydir` value.

**To reproduce**

Start the service without providing the `macaroonpath` or `faradaydir` flags and observe that the service starts as expected:
```
$ ./faraday --lnd.macaroonpath=/path/admin.macaroon --lnd.tlscertpath=/path/tls.cert --lnd.rpcserver=127.0.0.1:10009 --network=regtest
2024-08-14 11:24:28.612 [INF] FRDY: Generating TLS certificates...
2024-08-14 11:24:28.616 [INF] FRDY: Done generating TLS certificates
2024-08-14 11:24:28.713 [INF] FRPC: Starting gRPC listener
2024-08-14 11:24:28.713 [INF] FRPC: gRPC server listening on 127.0.0.1:8465
2024-08-14 11:24:28.713 [INF] FRPC: REST proxy disabled
```

Then attempt to start the service with the `maracroonpath` flag and not the `faradaydir` flag and observe the error:
```
./faraday --lnd.macaroonpath=/path/admin.macaroon --lnd.tlscertpath=/path/tls.cert --lnd.rpcserver=127.0.0.1:10009 --network=regtest --macaroonpath=/macpath                 
Error starting faraday: error validating config: faradaydir overwrites macaroonpath, please only set one value%  
```

**System information**

Faraday is running standalone, however, I have observed similar behavior with litd. I am using Faraday v0.2.13 cloned from the repo and compiled locally.

ellemouton:
1) Ensure that the `faradayDirSet` boolean is accurately set. Otherwise it is always true & so will always result in errors when the user provides their own macaroon path when they dont set the faraday dir.
2) For frcli, ensure that the tls cert & macaroon paths are only overridden with the correct network namespace if the user has not explicitly defined them

Fixes #191

### DIFF
--- a/cmd/frcli/utils.go
+++ b/cmd/frcli/utils.go
@@ -159,12 +159,23 @@ func extractPathArgs(ctx *cli.Context) (string, string, error) {
 	if faradayDir != faraday.FaradayDirBase ||
 		networkStr != faraday.DefaultNetwork {
 
-		tlsCertPath = filepath.Join(
-			faradayDir, networkStr, faraday.DefaultTLSCertFilename,
-		)
-		macPath = filepath.Join(
-			faradayDir, networkStr, faraday.DefaultMacaroonFilename,
-		)
+		// Only overwrite the tls cert path if the user has not
+		// explicitly defined it.
+		if !ctx.GlobalIsSet(tlsCertFlag.Name) {
+			tlsCertPath = filepath.Join(
+				faradayDir, networkStr,
+				faraday.DefaultTLSCertFilename,
+			)
+		}
+
+		// Only overwrite the macaroon path if the user has not
+		// explicitly defined it.
+		if !ctx.GlobalIsSet(macaroonPathFlag.Name) {
+			macPath = filepath.Join(
+				faradayDir, networkStr,
+				faraday.DefaultMacaroonFilename,
+			)
+		}
 	}
 
 	return tlsCertPath, macPath, nil
--- a/config.go
+++ b/config.go
@@ -192,6 +192,10 @@ func ValidateConfig(config *Config) error {
 	config.TLSKeyPath = lncfg.CleanAndExpandPath(config.TLSKeyPath)
 	config.MacaroonPath = lncfg.CleanAndExpandPath(config.MacaroonPath)
 
+	// Before adding the network namespace below, check if the user has
+	// overwritten the default faraday directory.
+	faradayDirSet := config.FaradayDir != FaradayDirBase
+
 	// Append the network type to faraday directory so they are "namespaced"
 	// per network.
 	config.FaradayDir = filepath.Join(config.FaradayDir, config.Network)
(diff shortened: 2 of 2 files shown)


cmd/frcli/utils.go
```diff
@@ -159,12 +159,23 @@ func extractPathArgs(ctx *cli.Context) (string, string, error) {
...
 
-		tlsCertPath = filepath.Join(
-			faradayDir, networkStr, faraday.DefaultTLSCertFilename,
-		)
-		macPath = filepath.Join(
-			faradayDir, networkStr, faraday.DefaultMacaroonFilename,
-		)
+		// Only overwrite the tls cert path if the user has not
+		// explicitly defined it.
+		if !ctx.GlobalIsSet(tlsCertFlag.Name) {
```
guggero, line 164: I guess the alternative way for detecting this is to compare the value against the default value (which I think is what we do in `lnd` and `litd`), which also works when using a config file. So if we ever add config file support to Faraday, this will likely break...

config.go
```diff
@@ -192,6 +192,10 @@ func ValidateConfig(config *Config) error {
 	config.TLSKeyPath = lncfg.CleanAndExpandPath(config.TLSKeyPath)
 	config.MacaroonPath = lncfg.CleanAndExpandPath(config.MacaroonPath)
 
+	// Before adding the network namespace below, check if the user has
+	// overwritten the default faraday directory.
+	faradayDirSet := config.FaradayDir != FaradayDirBase
```
guggero, line 197: Nice catch!

guggero (commented):
Thanks for the quick fix!

cmd/frcli/utils.go
ellemouton: ah - I thought config files are only for the long running daemons and not the CLIs right?

ellemouton: but I can copy the pattern used in LNDs `extractPathArgs` if that will be better?

guggero: Oh, you're absolutely right, I completely missed that part, didn't look at the file this change was made in... Okay, never mind then.

guggero (approved):
Sorry, forgot that I didn't approve yet. LGTM :tada:

ViktorTigerstrom (approved):
tACK LGTM :rocket:! Thanks for the fix :)

PR #194, opened 2024-08-19
frdrpc: convert frdrpc into a go module

bhandras:
This commit transforms the frdrpc subpackage into a versioned module. By doing so, projects that rely solely on the generated RPC code from the proto files will benefit from significantly reduced dependencies during compilation.

#### Pull Request Checklist


frdrpc/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/lightninglabs/faraday/frdrpc
+
+go 1.22.3
```
guggero, line 3: nit: when we're touching or creating all these `go.mod` files, can we please also sync the place we put the version? Either all at the top or the bottom of the file? Instinctively I'd say probably at the bottom...

go.mod
```diff
@@ -5,17 +5,18 @@ require (
...
 	github.com/lightninglabs/lndclient v0.17.4-1
 	github.com/lightningnetwork/lnd v0.17.4-beta
 	github.com/lightningnetwork/lnd/cert v1.2.2
 	github.com/lightningnetwork/lnd/kvdb v1.4.4
 	github.com/shopspring/decimal v1.2.0
 	github.com/stretchr/testify v1.8.4
 	github.com/urfave/cli v1.22.9
-	google.golang.org/grpc v1.59.0
-	google.golang.org/protobuf v1.31.0
+	google.golang.org/grpc v1.65.0
```
guggero, line 18: Hmm, I'm not sure if this update might bite us, since in lnd we have an older version. Or is the plan to update that as well?

tools/Dockerfile
```diff
@@ -1,17 +1,17 @@
...
 
 RUN cd /tmp \
   && mkdir -p /tmp/build/.cache \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
-  && go install -trimpath -tags=tools github.com/golangci/golangci-lint/cmd/golangci-lint \
-  && chmod -R 777 /tmp/build/
+  && go install -trimpath github.com/golangci/golangci-lint/cmd/golangci-lint \
+  && chmod -R 777 /tmp/build/ \
+  && git config --global --add safe.directory /build
```
guggero, line 15: I think the `ENV GOFLAGS="-buildvcs=false"` is the actual, preferred fix for the issue? Or did that no longer work?

tools/go.mod
```diff
@@ -1,9 +1,197 @@
 module github.com/lightninglabs/faraday/tools
 
-go 1.16
+go 1.21
```
guggero, line 3: We might as well bump this version too...

guggero (commented):
Thanks for the PR. A couple of minor things, otherwise looks good.

tools/Dockerfile
bhandras: Yeah, I think I was just simply adopting the `Dockerfile` that we use in other `lightninglabs` projects. Modified to use the suggested flag instead.

tools/go.mod
bhandras: Done.

bhandras (commented):
Thanks for the review @guggero! Attempted to address all your comments, PTAL 🙏

guggero (approved):
LGTM :tada:

PR #202, opened 2025-01-15
Only fetch onchain transactions from LND within start_time and end_time range

nordbjorn:
Ensure that the `NodeAudit` RPC makes use of the provided `start_time` and `end_time` arguments, in the sense that Faraday will only query the LND backend for onchain transactions using a block range which corresponds to the `start_time` and `end_time` (with a small buffer).

## Motivation and Context

The `NodeAudit` request can currently fail for old LND nodes, with a lot of historical onchain activity. Even if the caller provides a small time range to produce the report for. This is because currently all onchain transactions are fetched from the LND backend, and then filtered in memory based on the given time range.

This PR aims to alleviate most of the problem by ensuring Faraday does not need to query the entire history of onchain transactions when the caller provides a time range for the report.

## Implementation Notes

I did not find any obvious way to resolve a timestamp to a block height, which needs to be provided for the queries to the LND backend. For now I chose an approach where we do a binary search on the entire blockchain to find the block height just before the provided timestamp.

This means we need to look up several block headers for each timestamp we want to resolve. I think this approach is fine for now, and in this context, considering that this RPC is a slow and heavy one anyways. Worst case is that we need to fetch ~20 block headers per timestamp we look up.

## Future improvements

1. If we want to optimize the timestamp to block height resolving more, we could make use of this public API endpoint: https://mempool.space/docs/api/rest#get-block-timestamp
This endpoint seems to do a lookup from an internal DB (see [here](https://github.com/mempool/mempool/blob/d27bb7e156dfce8d5f56c2bf01e6e07b68715ce8/backend/src/repositories/BlocksRepository.ts#L563-L586)), which is likely more performant. In that case the current binary search mechanism could serve as only a fallback for if the mempool.space API is unavailable

2. We can also make use of pagination when querying onchain activity from the LND backend. This seems to require changes in the [lndclient](https://github.com/lightninglabs/lndclient) repository as well, to ensure that the `GetTransactions` RPC can be called with a `index_offset` and `max_transactions`, like described [here](https://lightning.engineering/api-docs/api/lnd/lightning/get-transactions/#grpc)

#### Pull Request Checklist


accounting/config.go
```diff
@@ -109,6 +111,58 @@ func NewOnChainConfig(ctx context.Context, lnd lndclient.LndServices, startTime,
 		}
 	}
 
+	// Start end end height can initially be set to 0, meaning we will query
```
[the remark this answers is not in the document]
nordbjorn: Thanks for the review @matheusavi !

I pushed a commit with the suggested fixes, as well as some minor styling improvements and log level adjustments I found myself. See here: a01826c1080ca3bdae881c21096269c48ee37380

accounting/config.go
```diff
@@ -109,6 +111,58 @@ func NewOnChainConfig(ctx context.Context, lnd lndclient.LndServices, startTime,
...
+
+		if startHeight > endHeight {
+			log.Errorf("Start height: %v is greater than end height: %v, "+
+				"setting both to 0", startHeight, endHeight)
+			startHeight = 0
+			endHeight = 0
+		}
+	}
+
+	log.Infof("Using startheight: %v endheight: %v while querying onchain"+
```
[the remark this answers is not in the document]
nordbjorn: a01826c1080ca3bdae881c21096269c48ee37380

accounting/config.go
```diff
@@ -109,6 +111,58 @@ func NewOnChainConfig(ctx context.Context, lnd lndclient.LndServices, startTime,
...
+			// and miss any transaction
+			endHeight += 3
+		}
+
+		if startHeight > endHeight {
+			log.Errorf("Start height: %v is greater than end height: %v, "+
+				"setting both to 0", startHeight, endHeight)
+			startHeight = 0
+			endHeight = 0
+		}
```
guggero, line 160: Sounds like we could extract this into a function as well, that's then passed in as `blockRangeLookup func(start, end time.Time) (uint32, uint32, error),`.

frdrpcserver/node_audit.go
```diff
@@ -84,8 +86,20 @@ func parseNodeAuditRequest(ctx context.Context, cfg *Config,
...
 	}
 
+	var blockHeightLookup func(targetTimestamp time.Time) (uint32, error)
+
+	// If a time range is set, we will use a block height lookup function
+	// to find the block heights for the start and end time.
+	timeRangeSet := req.StartTime > 0 || req.EndTime > 0
+	if timeRangeSet {
+		blockHeightLookup = func(targetTimestamp time.Time) (uint32, error) {
+			return findFirstBlockBeforeTimestamp(ctx, cfg.Lnd, info.BlockHeight,
```
guggero, line 96: nit: see formatting guideline for wrapping long function calls.

guggero (commented):
Thanks for the improvement!
Can you please squash the commits?

accounting/config.go
```diff
@@ -109,6 +111,58 @@ func NewOnChainConfig(ctx context.Context, lnd lndclient.LndServices, startTime,
 		}
 	}
 
+	// Start and end height can initially be set to 0, meaning we will query
+	// for all onchain history
```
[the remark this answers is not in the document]
nordbjorn: Thanks for the review and feedback!

I reviewed this doc and updated this PR to try and ensure it abides by these rules.

I also squashed the git history.

accounting/config.go
nordbjorn: Good idea. I refactored according to this now.

I also realized that the buffer of 3 blocks in either direction might not be enough in all circumstances. Since the rules for the block timestamp are somewhat loose, I think this could result in potentially fetching too few blocks from the backend. So I updated the logic here so that we apply a 24h buffer on the timestamps when resolving to a block height. I think that should be enough even in extreme network conditions. What do you think?

frdrpcserver/node_audit.go
nordbjorn: Thanks, this is also fixed now

frdrpcserver/node_audit.go
```diff
@@ -278,3 +293,133 @@ func rpcEntryType(t accounting.EntryType) (frdrpc.EntryType, error) {
...
+func resolveBlockHeightRange(ctx context.Context,
+	lndClient lndclient.LndServices, latestHeight uint32,
+	startTime, endTime time.Time) (uint32, uint32, error) {
+
+	// Since Bitcoin blocks are not guaranteed to be completely ordered
+	// by timestamp, and the timestamps can be manipulated by miners within a
+	// certain range, we will apply a buffer on the time which we use to
+	// find the block height. This should ensure we use a low enough
+	// height to fetch all relevant transactions following the start time.
+	bufferedStartTime := startTime.Add(time.Hour * -24)
```
guggero, line 311: nit: probably would make sense to extract this range value into a named variable at the start of the file.

guggero (approved):
tACK, LGTM :tada: 

Thanks a lot for the fix!

frdrpcserver/node_audit.go
nordbjorn: Thanks for the review!

Agreed, I refactored according to this now.

I simply squashed this change into the existing commit.

bhandras (approved):
LGTM, thank you! 🎉

PR #218, opened 2025-05-02
fiat: add support for using Coinbase to fetch hourly/daily prices

bhandras:
With this commit we add Coinbase to the existing fiat sources.

#### Pull Request Checklist

### DIFF
--- a/fiat/coinbase_api.go
+++ b/fiat/coinbase_api.go
@@ -1,0 +1,161 @@
+package fiat
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/shopspring/decimal"
+)
+
+const (
+	coinbaseHistoryAPI  = "https://api.exchange.coinbase.com/products/%s/candles"
+	coinbaseDefaultPair = "BTC-USD"
+	coinbaseCandleCap   = 300   // max buckets.
+	coinbaseGranHourSec = 3600  // 1‑hour buckets.
+	coinbaseGranDaySec  = 86400 // 1‑day  buckets.
+	coinbaseDefaultCurr = "USD"
+)
+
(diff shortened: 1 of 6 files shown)


fiat/coinbase_api.go
```diff
@@ -0,0 +1,157 @@
...
+	resp, err := cl.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	return io.ReadAll(resp.Body)
+}
+
+// paarseCoinbaseData parses the JSON response from Coinbase's candles endpoint.
```
hieblmi, line 66: typo: paarse

fiat/coinbase_api.go
```diff
@@ -0,0 +1,157 @@
...
+//	]
+func parseCoinbaseData(data []byte) ([]*Price, error) {
+	var raw [][]float64
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return nil, err
+	}
+
+	prices := make([]*Price, 0, len(raw))
+	for _, c := range raw {
+		if len(c) < 5 {
```
hieblmi, line 108: should this be `!= 5`, or are there only cases with less than 5?

hieblmi (commented):
Just a small drive by review.

fiat/coinbase_api.go
bhandras: Yeah good point. According to their docs: "Historical rate data may be incomplete. No data is published for intervals where there are no ticks.",  so i'd say we should allow for rows where the data is "inclomplete" ie. likely empty. It's hard to say if they'd change their API's output in a non-backward compatible way,  but I think it's safe to expect that they'd like introduce any breaking changes under a separate url path, eg. `/v2/` etc. Added a comment clarifying incompleteness.

hieblmi (approved):
LGTM!

PR #225, opened 2025-10-02
accounting: add error messages for errors

starius:
Replaced `return err` cases with wrapped errors to attach the info which may be useful when debugging.

#### Pull Request Checklist


bhandras (approved):
Very nice! Thank you 🎉

PR #232, opened 2026-03-04
mod+itest: update to lnd v0.20.1

bitromortac:
Updates itests from v0.15 to v0.20.1.

Todo: 
- [x] find out why the channel open fee went down:
This is because of this change https://github.com/lightningnetwork/lnd/pull/9257, where the previous regtest fee estimation had a fee of 50sat/vbyte, but now the fallback is 25sat/vbyte.


go.mod
```diff
@@ -4,13 +4,13 @@ require (
...
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0
-	github.com/btcsuite/btclog/v2 v2.0.1-0.20250110154127-3ae4bf1cb318
+	github.com/btcsuite/btclog/v2 v2.0.1-0.20250728225537-6090e87c6c5b
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.22.0
 	github.com/jarcoal/httpmock v1.4.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/lightninglabs/faraday/frdrpc v1.0.1
-	github.com/lightninglabs/lndclient v0.19.0-7
-	github.com/lightningnetwork/lnd v0.19.0-beta
+	github.com/lightninglabs/lndclient v0.20.0-6
```
ViktorT-11: Why don't we update to `lndclient v0.20.0-7` while we're at it :)?

itest/nodereport_test.go
```diff
@@ -85,7 +85,7 @@ func TestNodeAudit(t *testing.T) {
 	feeRef := accounting.FeeReference(aliceChannel.Hash.String())
 	expected[feeRef] = expectedReport{
 		eventType: frdrpc.EntryType_CHANNEL_OPEN_FEE,
-		amount:    lnwire.MilliSatoshi(8237 * 1000),
+		amount:    lnwire.MilliSatoshi(4118 * 1000),
```
ViktorT-11, line 88: I think it'd be helpful if you explain in the commit message why these amount changes here are needed. Also, as I believe this test will fail after only applying the first commit (might be wrong here though given the itest `Dockerfile` bump), maybe it'd make most sense to squash these commits together :)?

ViktorT-11 (commented):
Generally looks good :fire:! Just leaving 2 small suggestions.

go.mod
bitromortac: done :)

itest/nodereport_test.go
bitromortac: I think they succeed independently since the lnd version is defined by the version defined in the docker file as you noted

ViktorT-11 (approved):
Thanks, LGTM :rocket:!

PR #233, opened 2026-03-09
mod: use chan events lnd and lndclient

bitromortac:
Updates the `faraday-forwarding-ability` branch to use latest lnd/lndclient for being able to receive channel update events. We'll add more functionality later using this side branch.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,8 +9,8 @@ require (
 	github.com/jarcoal/httpmock v1.4.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/lightninglabs/faraday/frdrpc v1.0.1
-	github.com/lightninglabs/lndclient v0.20.0-7
-	github.com/lightningnetwork/lnd v0.20.1-beta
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -193,4 +193,4 @@ require (
 // allows us to specify that as an option.
 replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.30.0-hex-display
 
-go 1.24.11
+go 1.25.5
```
ViktorT-11, line 196: I think we should either mention the `Golang` in the commit message + title, or bump it through a separate commit.

Additionally, the commit message should also make it clean that `lnd` + `lndclient` is bumped.

ViktorT-11 (commented):
Generally looks good once the commit message issue has been adressed :)

go.mod
bitromortac: Done, I separated them into two commits, I think it's cleaner that way 👍

ViktorT-11 (approved):
LGTM

PR #234, opened 2026-03-12
faraday+frdrpcserver: make faraday an active component

bitromortac:
This extracts server lifecycle management from `frdrpcserver.RPCServer` into a new top-level `Faraday` struct, turning `RPCServer` into a thin handler wrapper. Faraday now owns gRPC/REST servers, TLS, macaroons, and the lnd connection, enabling it to run as a first-class subserver in litd rather than just an RPC handler. This is needed to later watch for channel event updates, as the alternative would have been to let the goroutine run in the RPC subserver, which is a violation of concerns.

See [architecture diagram](https://github.com/bitromortac/faraday/blob/74e4a7c7926ef77b6acaa3495042b38a4803489c/.reviews/architecture.md) for a before/after overview.

### DIFF
--- a/faraday.go
+++ b/faraday.go
@@ -2,18 +2,60 @@
 package faraday
 
 import (
+	"context"
+	"crypto/tls"
 	"fmt"
+	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
 
+	proxy "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/jessevdk/go-flags"
 	"github.com/lightninglabs/faraday/chain"
+	"github.com/lightninglabs/faraday/frdrpc"
 	"github.com/lightninglabs/faraday/frdrpcserver"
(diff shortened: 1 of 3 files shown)


ViktorT-11 (commented):
I think this mainly looks good, but this PR is quite hard to follow.

I think the reason for that is mainly that the first commit both mixes refactors & new additions in the same commit. The commit message for the commit is also very sparse.

Could potentially structure it like this (even though this might be a bit hard):
* Commit 1: Refactors only and move code from the `frdrpcserver` to the & `faraday` package.
* Commit 2: Introduction of the new additions that this PR also adds.

That'd be a much cleaner and easier way to follow what is actually changed here and what is just moved between the packages.

bitromortac:
> I think this mainly looks good, but this PR is quite hard to follow.

Right, sorry about that, I tried to split it up. I think it should be better now, though I added some other fixes that came up in self-review.

ellemouton (approved):
nice!

just the last commit should be removed :)

ViktorT-11 (approved):
Thanks for the changes, LGTM :fire:!

PR #235, opened 2026-03-24
db: introduce sqldbv2 scaffolding

bitromortac:
Based on https://github.com/lightninglabs/faraday/pull/234.

This PR introduces a SQL database to Faraday, which will later be used to track channel state changes.

Uses the exact sqldbv2 version of https://github.com/lightninglabs/lightning-terminal/tree/sql-migration-base and the custom migration library. The scaffolding was copied from this (already reviewed) branch, which should make it easy to compare against.

Corresponding lnd PR: https://github.com/lightningnetwork/lnd/pull/10175


bitromortac:
Updated to latest sqldbv2 in lnd side branch.

Makefile
```diff
@@ -137,6 +137,14 @@ rpc-js-compile:
 	@$(call print, "Compiling JSON/WASM stubs.")
 	GOOS=js GOARCH=wasm $(GOBUILD) $(PKG)/frdrpc
 
+sqlc:
+	@$(call print, "Generating sql models and queries in Go")
+	./scripts/gen_sqlc_docker.sh
+
+sqlc-check: sqlc
```
ViktorT-11, line 144: Should add this check to the CI workflow.

make/testing_flags.mk
```diff
@@ -1,6 +1,18 @@
...
+ifeq ($(dbbackend),postgres)
+DEV_TAGS += test_db_postgres
+else
+DEV_TAGS += test_db_sqlite
+endif
+
+# Add any additional tags that are passed in to make.
+ifneq ($(tags),)
+DEV_TAGS += ${tags}
+endif
```
ViktorT-11, line 15: Nit: Could potentially have this part in a separate commit.

make/testing_flags.mk
```diff
@@ -1,6 +1,18 @@
 TEST_FLAGS =
```
ViktorT-11, line 1: Should have a `DEV_TAGS = dev` here?

db/sqlerrors.go
```diff
@@ -0,0 +1,198 @@
...
+
+var (
+	// ErrRetriesExceeded is returned when a transaction is retried more
+	// than the max allowed valued without a success.
+	ErrRetriesExceeded = errors.New("db tx retries exceeded")
+)
+
+// MapSQLError attempts to interpret a given error as a database agnostic SQL
+// error.
+func MapSQLError(err error) error {
```
ViktorT-11: Hmmm I think we can probably use the `sqldb/v2` version of this instead of having this file! Worth looking into. We should probably do the same in `litd`.

db/postgres.go
```diff
@@ -0,0 +1,72 @@
...
+	SkipMigrations     bool          `long:"skipmigrations" description:"Skip applying migrations on startup."`
+	Host               string        `long:"host" description:"Database server hostname."`
+	Port               int           `long:"port" description:"Database server port."`
+	User               string        `long:"user" description:"Database user."`
+	Password           string        `long:"password" description:"Database user's password."`
+	DBName             string        `long:"dbname" description:"Database name to use."`
+	MaxOpenConnections int           `long:"maxconnections" description:"Max open connections to keep alive to the database server."`
+	MaxIdleConnections int           `long:"maxidleconnections" description:"Max number of idle connections to keep in the connection pool."`
+	ConnMaxLifetime    time.Duration `long:"connmaxlifetime" description:"Max amount of time a connection can be reused for before it is closed. Valid time units are {s, m, h}."`
+	ConnMaxIdleTime    time.Duration `long:"connmaxidletime" description:"Max amount of time a connection can be idle for before it is closed. Valid time units are {s, m, h}."`
```
ViktorT-11: Hmmm we should look into if we can use the `sqldb/v2` variants for these. We should check this in `litd` as well.

db/sqlc/db_custom.go
```diff
@@ -0,0 +1,41 @@
+// Package sqlc provides a set of custom database queries and utilities
```
ViktorT-11: Not sure if we should add this file before we actually need custom queries.

ViktorT-11 (commented):
Nice, this will definitely be useful. Leaving a few comments below 😃

faraday.go
```diff
@@ -26,6 +75,479 @@ var MinLndVersion = &verrpc.Version{
...
+			defer f.wg.Done()
+			err := f.restServer.Serve(restListener)
+			// ErrServerClosed is always returned when the proxy is
+			// shut down, so don't log it.
+			if err != nil && err != http.ErrServerClosed {
+				log.Error(err)
+			}
+		}()
+	} else {
+		log.Infof("REST proxy disabled")
```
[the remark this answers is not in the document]
bitromortac: These comments are wrong

db/migrations.go
```diff
@@ -0,0 +1,10 @@
+package db
+
+const (
+	// LatestMigrationVersion is the latest migration version of the
+	// database. This is used to implement downgrade protection for the
+	// daemon.
+	//
+	// NOTE: This MUST be updated when a new migration is added.
+	LatestMigrationVersion = 1
```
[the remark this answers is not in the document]
bitromortac: Will add a test in a later PR that checks this.

Makefile
bitromortac: Right, I'll add that when we have the migrations since it would fail at the moment: `sed: can't read db/sqlc/migrations/*.up.sql: No such file or directory`

db/sqlerrors.go
bitromortac: Nice :tada:, removed that.

db/postgres.go
bitromortac: I am using the package config struct instead, thank you for the hint :pray:

db/sqlc/db_custom.go
bitromortac: Agree, removed it

bitromortac:
Thanks for the great review @ViktorT-11, I could remove a lot of code :tada:

Makefile
ViktorT-11: Makes sense :)

ViktorT-11 (approved):
Thanks for the changes, LGTM :fire:!

PR #236, opened 2026-03-27
db: implement sql tables and store for channel events

bitromortac:
Based on https://github.com/lightninglabs/faraday/pull/235.

Implements tables to store channel events. Events can either be records of channels becoming on/offline, or updates of channel balances.


db/sqlc/migrations/000001_chanevents.up.sql
```diff
@@ -0,0 +1,46 @@
...
+    event_type SMALLINT NOT NULL,
+    -- The time the event occurred.
+    timestamp TIMESTAMP NOT NULL,
+    -- The local balance of the channel at the time of the event.
+    -- This is only populated for balance update events.
+    local_balance_sat BIGINT CHECK (local_balance_sat >= 0),
+    -- The remote balance of the channel at the time of the event.
+    -- This is only populated for balance update events.
+    remote_balance_sat BIGINT CHECK (remote_balance_sat >= 0),
+    CHECK (event_type BETWEEN 1 AND 3)
```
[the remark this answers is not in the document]
ViktorT-11: I think we should consider if this check is necessary, as it'll require us to do a migration whenever there's a new `event_type`?

db/migrations_test.go
```diff
@@ -0,0 +1,40 @@
...
+		require.NoError(t, err)
+
+		if uint(v) > maxVersion {
+			maxVersion = uint(v)
+		}
+	}
+
+	require.EqualValues(
+		t, maxVersion, LatestMigrationVersion,
+		"LatestMigrationVersion is out of date, update db/migrations.go",
```
ViktorT-11: Nit: Line length

db/sqlc/migrations/000001_chanevents.up.sql
```diff
@@ -0,0 +1,46 @@
...
+    id INTEGER PRIMARY KEY,
+    -- The public key of the peer.
+    pubkey TEXT NOT NULL UNIQUE
+);
+
+-- The channels table stores all the channels that we have with our peers.
+CREATE TABLE IF NOT EXISTS channels (
+    -- The auto incrementing primary key.
+    id INTEGER PRIMARY KEY,
+    -- The channel point, as a 'txid:output_index' string.
```
ViktorT-11, line 13: This line is changed in the last commit for some reason 🤓

db/sqlc/db_custom.go
```diff
@@ -0,0 +1,34 @@
+// Package sqlc provides a set of custom database queries and utilities
```
ViktorT-11, line 1: This file is reintroduced again for reason 😁?

chanevents/store.go
```diff
@@ -0,0 +1,268 @@
...
+			ChannelID:   channelID,
+			Timestamp:   startTime.UTC(),
+			Timestamp_2: endTime.UTC(),
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get channel events: %w", err)
+	}
+
+	events := make([]*ChannelEvent, len(dbEvents))
```
ViktorT-11, line 243: Could consider if it's worth using a marshal function here similar to how we do it in `litd` when the db results are complex. But this is non-blocking and up to you.

chanevents/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+	// Adding the same peer again violates the unique constraint.
+	_, err = store.AddPeer(ctx, peer.PubKey)
+	require.Error(t, err)
+
+	dbPeer, err := store.GetPeer(ctx, "non_existent_pubkey")
+	require.ErrorIs(t, err, errUnknownPeer)
+	require.Nil(t, dbPeer)
+
+	// Get the peer and assert it is the same.
+	dbPeer, err = store.GetPeer(ctx, peer.PubKey)
```
ViktorT-11, line 49: nit: could attempt to fetch a non existent peer.

chanevents/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+	dbChannel, err := store.GetChannel(ctx, "non-existent-chan-point")
+	require.ErrorIs(t, err, errUnknownChannel)
+	require.Nil(t, dbChannel)
+
+	// Get the channel and assert it is the same.
+	dbChannel, err = store.GetChannel(ctx, testChanPoint1)
+	require.NoError(t, err)
+	require.Equal(t, testChanPoint1, dbChannel.ChannelPoint)
+	require.Equal(t, testShortChanID1, dbChannel.ShortChannelID)
+	require.Equal(t, peerID, dbChannel.PeerID)
```
ViktorT-11, line 78: nit: could potentially create a channel comparison helper function instead.

chanevents/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.Equal(t, testTime.Unix(), events[0].Timestamp.Unix())
+	require.True(t, events[0].LocalBalance.IsNone())
+	require.True(t, events[0].RemoteBalance.IsNone())
+
+	require.Equal(t, updateEvent.EventType, events[1].EventType)
+	require.Equal(
+		t, testTime.Add(time.Second).Unix(), events[1].Timestamp.Unix(),
+	)
+	require.Equal(t, updateEvent.LocalBalance, events[1].LocalBalance)
+	require.Equal(t, updateEvent.RemoteBalance, events[1].RemoteBalance)
```
ViktorT-11, line 127: Same here, could use a equal helper. Non of these are blocking, and optional though :)

ViktorT-11 (commented):
Looks great, just leaving a few suggestions :fire:!

db/sqlc/migrations/000001_chanevents.up.sql
bitromortac: Rebase error :)

db/sqlc/db_custom.go
bitromortac: Ah yes, we need `NewForType`. Moved it to the first commit as well as it fits to the commit message prefix.

chanevents/store.go
bitromortac: Done!

chanevents/store_test.go
bitromortac: this is done a few lines above, right?

chanevents/store_test.go
bitromortac: I skipped this, should we do more comparisons I'll add it, also for events.

chanevents/store_test.go
ViktorT-11: My bad.

ViktorT-11 (approved):
Nice! LGTM :fire:!

PR #237, opened 2026-04-02
chanevents+faraday: add chan events store and start

bitromortac:
Adds the channel event monitor and collects channel update events.

### DIFF
--- a/chanevents/monitor.go
+++ b/chanevents/monitor.go
@@ -1,0 +1,508 @@
+package chanevents
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightningnetwork/lnd/fn/v2"
+	"github.com/lightningnetwork/lnd/routing/route"
+)
+
+const (
+	// retryInterval is the time to wait before retrying after a
+	// transient error or while waiting for lnd to become ready.
+	retryInterval = 5 * time.Second
+)
+
+var (
+	// errMonitorAlreadyStarted is returned when the monitor is already
+	// started.
+	errMonitorAlreadyStarted = errors.New("monitor already started")
+
+	// errMonitorNotStarted is returned when the monitor is not started.
+	errMonitorNotStarted = errors.New("monitor not started")
+)
+
+// Monitor is an active component that listens to LND channel events and records
+// them in the database.
+type Monitor struct {
+	started atomic.Bool
+
+	// lnd is the lnd client that the monitor will use to subscribe to
+	// channel events.
+	lnd lndclient.LightningClient
+
+	// store is the channel events store that the monitor will use to record
+	// channel events.
+	store *Store
+
+	wg   sync.WaitGroup
+	quit chan struct{}
+}
+
+// NewMonitor creates a new channel events monitor.
+func NewMonitor(lnd lndclient.LightningClient, store *Store) *Monitor {
+	return &Monitor{
+		lnd:   lnd,
+		store: store,
+		quit:  make(chan struct{}),
+	}
+}
+
+// Start starts the channel events monitor.
+func (m *Monitor) Start(ctx context.Context) error {
+	if !m.started.CompareAndSwap(false, true) {
+		return errMonitorAlreadyStarted
+	}
+
+	log.Info("Starting channel events monitor")
+
+	m.quit = make(chan struct{})
+
+	m.wg.Add(1)
+	go m.monitorLoop(ctx)
+
+	return nil
+}
+
+// Stop stops the channel events monitor.
+func (m *Monitor) Stop() error {
+	if !m.started.CompareAndSwap(true, false) {
+		return errMonitorNotStarted
+	}
+
+	log.Info("Stopping channel events monitor")
+
+	close(m.quit)
+	m.wg.Wait()
+
+	return nil
+}
+
+// monitorLoop is the main loop of the channel events monitor. It waits for lnd
+// to be fully synced, performs an initial state sync, and then subscribes to
+// channel events. If the subscription fails or the stream breaks, it retries
+// from the beginning.
+func (m *Monitor) monitorLoop(ctx context.Context) {
+	defer m.wg.Done()
+
+	log.Info("Channel events monitor starting")
+
+	for {
+		// Wait for lnd to be synced to chain, retrying on RPC errors.
+		if !m.waitForReady(ctx) {
+			return
+		}
+
+		// Initial state sync.
+		if err := m.initialSync(ctx); err != nil {
+			log.Errorf("Error during initial sync: %v", err)
+		}
+
+		// Subscribe and consume events until the stream breaks or an
+		// error occurs.
+		if !m.subscribe(ctx) {
+			return
+		}
+
+		// Stream broke, wait before reconnecting.
+		log.Infof("Reconnecting channel event subscription...")
+
+		select {
+		case <-time.After(retryInterval):
+		case <-m.quit:
+			return
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// waitForReady polls lnd's GetInfo until it reports SyncedToChain. It retries
+// on transient RPC errors. It returns true when lnd is ready, or false if the
+// monitor is shutting down.
+func (m *Monitor) waitForReady(ctx context.Context) bool {
+	for {
+		info, err := m.lnd.GetInfo(ctx)
+		if err != nil {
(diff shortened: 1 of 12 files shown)


log.go
```diff
@@ -3,6 +3,7 @@ package faraday
 import (
 	"github.com/btcsuite/btclog/v2"
 	"github.com/lightninglabs/faraday/accounting"
+	"github.com/lightninglabs/faraday/chanevents"
```
bitromortac, line 6: will remove the root logger mention from the commit message

chanevents/monitor.go
```diff
@@ -0,0 +1,418 @@
...
+				log.Errorf("error handling channel event: %v", err)
+			}
+
+		case err, ok := <-errChan:
+			if !ok {
+				log.Info("Channel event error stream closed")
+				return
+			}
+			log.Errorf("error from channel event subscription: %v", err)
+			return
```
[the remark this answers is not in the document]
bitromortac: This will be addressed in a later PR.

chanevents/monitor.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// Initial state sync.
+	if err := m.initialSync(ctx); err != nil {
+		log.Errorf("error during initial sync: %v", err)
+		// We'll continue anyway, maybe the subscription will work.
+	}
+
+	// Subscribe to channel events.
+	// TODO: Add reconnection logic.
+	eventChan, errChan, err := m.lnd.SubscribeChannelEvents(ctx)
```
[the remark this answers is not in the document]
bitromortac: Would address this in a later PR.

chanevents/monitor.go
```diff
@@ -0,0 +1,418 @@
...
+			LocalBalance:  fn.Some(channel.LocalBalance),
+			RemoteBalance: fn.Some(channel.RemoteBalance),
+		}); err != nil {
+			log.Errorf("error adding event for channel %s: %v",
+				channel.ChannelPoint, err)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
bitromortac: This is true as well, but I think we want to address this later. An action I took was to add a isSync row to signal that the event was recorded during startup, to distinguish real events from startup sync.

chanevents/store_test.go
```diff
@@ -22,6 +22,22 @@ var (
 	testTime = time.Unix(1, 0)
 )
 
+// requireEqualEvent asserts that a retrieved event matches the expected values,
+// comparing only the fields that are set before insertion (ignoring the
+// auto-assigned ID).
+func requireEqualEvent(t *testing.T, expected *ChannelEvent,
```
ViktorT-11, line 28: 😃

db/sqlc/migrations/000001_chanevents.up.sql
```diff
@@ -35,7 +35,10 @@ CREATE TABLE IF NOT EXISTS channel_events (
     local_balance_sat BIGINT CHECK (local_balance_sat >= 0),
     -- The remote balance of the channel at the time of the event.
     -- This is only populated for balance update events.
-    remote_balance_sat BIGINT CHECK (remote_balance_sat >= 0)
+    remote_balance_sat BIGINT CHECK (remote_balance_sat >= 0),
```
ViktorT-11, line 38: We see this as ok instead of doing a separate migration as the sql `db` isn't deployed anywhere right? If yes, I agree with this approach instead of already adding a bunch of migrations.

config.go
```diff
@@ -396,3 +426,99 @@ func loadCertWithCreate(cfg *Config) (tls.Certificate, *x509.Certificate,
...
+	// ChanEventsStore is used to watch for channel events.
+	ChanEventsStore *chanevents.Store
+
+	// closeFns holds various callbacks that can be used to close any open
+	// stores in the stores struct.
+	closeFns map[string]func() error
+}
+
+// NewStores creates a new stores instance based on the chosen database backend.
+func NewStores(cfg Config, clock clock.Clock) (*stores, error) {
```
ViktorT-11, line 441: Due to https://github.com/lightninglabs/faraday/pull/237/changes#r3038797802, perhaps we should put this (i.e. creation of the real store) behind a `dev` tag until we know for sure that we have what we need for the functionalities we want to support for now? As it's merged into a side branch, I guess you could motivate that as not needed though, but then we should be careful not to merge this to `master` before we're sure that we have everything we need.

chanevents/monitor.go
```diff
@@ -0,0 +1,431 @@
...
+		if err := m.store.AddChannelEvent(ctx, &ChannelEvent{
+			ChannelID:     dbChan.ID,
+			EventType:     EventTypeUpdate,
+			LocalBalance:  fn.Some(channel.LocalBalance),
+			RemoteBalance: fn.Some(channel.RemoteBalance),
+			IsSync:        true,
+		}); err != nil {
+			log.Errorf("error adding event for channel %s: %v",
+				channel.ChannelPoint, err)
+		}
```
ViktorT-11, line 324: Do we want to do this in 2 separate channel events here? If yes, maybe add a comment explaining why.

chanevents/monitor.go
```diff
@@ -0,0 +1,431 @@
...
+	return nil
+}
+
+// monitorLoop is the main loop of the channel events monitor.
+func (m *Monitor) monitorLoop(ctx context.Context) {
+	defer m.wg.Done()
+
+	log.Info("Channel events monitor starting")
+
+	// TODO: Wait for lnd to be fully synced?
```
ViktorT-11: This can be an issue if `litd` is in `remote` mode, so you should consider it. In `integrated` mode, we will not start `faraday` until `lnd` is synced (at least with the current logic).

faraday.go
```diff
@@ -442,6 +453,23 @@ func (f *Faraday) Stop() error {
 	// can complete cleanly.
 	f.wg.Wait()
 
+	if f.ctxCancel != nil {
```
ViktorT-11: Should the context be canceled prior to stopping the `monitor` which uses the context?

chanevents/monitor.go
```diff
@@ -0,0 +1,431 @@
...
+		if err := m.store.AddChannelEvent(ctx, &ChannelEvent{
+			ChannelID: dbChan.ID,
+			EventType: eventType,
+			IsSync:    true,
+		}); err != nil {
+			log.Errorf("error adding event for channel %s: %v",
+				channel.ChannelPoint, err)
+		}
+
+		if err := m.store.AddChannelEvent(ctx, &ChannelEvent{
```
ViktorT-11, line 315: Just want to ask regarding how we handle events which happen when `faraday` is offline, but `lnd` isn't.
  
Is the idea here that if `faraday` is offline while an `lnd` is online (hence `lnd.SubscribeChannelEvents` hasn't been executed) and a channel event occurs, that this will always ensure such events gets accounted for by this update.

I think should should add docs that explains how this is accounted for, as that is really valuable info to understand.

ViktorT-11 (commented):
Generally looks great! Just adding a few comments (mainly questions).

db/sqlc/migrations/000001_chanevents.up.sql
bitromortac: right, it's in a side branch so I'd give us the freedom to do so

config.go
bitromortac: Yes, I think we'll only merge once we have full interop and finalized the testing.

chanevents/monitor.go
bitromortac: Yes, because the events serve different purposes, one is for uptime calculation, the other for balance availability. Added a comment

faraday.go
bitromortac: Addressed this :pray:

chanevents/monitor.go
bitromortac: Right, the initial sync can't recover individual events that happened while faraday was offline, but it captures the latest state as a baseline (current balances, online/offline status), so we need to accept that there may be gaps in there, which can be circumvented by running faraday+lnd in integrated mode. At least there's a marker `IsSync` to see that those updates come from the sync. Added a comment.

chanevents/monitor.go
bitromortac: I added a retry loop, which will improve situations where we have transient errors (no reconnection is performed though).

chanevents/monitor.go
```diff
@@ -0,0 +1,508 @@
...
+	log.Info("Channel events monitor starting")
+
+	for {
+		// Wait for lnd to be synced to chain, retrying on RPC errors.
+		if !m.waitForReady(ctx) {
+			return
+		}
+
+		// Initial state sync.
+		if err := m.initialSync(ctx); err != nil {
```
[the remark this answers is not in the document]
bitromortac: Resolved by only syncing once.

PR #239, opened 2026-04-28
chanevents: implement GetChannelEvents rpc

bitromortac:
Exposes recorded channel events via RPC and adds an itest to test correct channel update recording.


litbot-9000:
@bitromortac, remember to re-request review from reviewers when ready

frdrpc/faraday.proto
```diff
@@ -596,3 +601,60 @@ message CloseReportResponse {
...
+    The channel point of the channel to get events for, formatted txid:outpoint.
+    */
+    string chan_point = 1;
+
+    /*
+    The start time of the query range, expressed as unix epoch offset in
+    seconds. Inclusive. Also serves as the pagination cursor: callers retrieving
+    more than max_events should re-query with start_time set to the timestamp
+    of the last event returned by the previous call.
+    */
```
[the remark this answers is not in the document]
bitromortac: Resolved via design change to index-based pagination.

db/sqlc/queries/chanevents.sql
```diff
@@ -21,5 +21,9 @@ INSERT INTO channel_events (
 
 -- name: GetChannelEvents :many
 SELECT * FROM channel_events
-WHERE channel_id = $1 AND timestamp >= $2 AND timestamp < $3
-ORDER BY timestamp ASC, id ASC;
+WHERE channel_id = $1
+  AND id > $2
```
ViktorT-11, line 25: In general this PR looks great, so I think this is my main comment for the PR:
Is it for performance reasons or that you think it's a better UI that you use the `lastId` approach over using an offset (i.e. something like `LIMIT $5 OFFSET $6`)? The approach choosen effects the rest of the PR.

I think the offset approach is much more common, and is what we use in `litd` for `actions` and I think potentially more natural for a user at it leaks no internals of how the `ChannelEvent` is stored (i.e. the `id`), and is also independent of the store used (i.e. not requiring that an incremental `id` is used).

Would therefore be interesting to hear your motivation behind this approach, which I do see some pros with as well!

itest/channel_events_test.go
```diff
@@ -0,0 +1,132 @@
...
+		)
+		require.NoError(c.t, err, "could not get paginated events")
+
+		paged = append(paged, page.Events...)
+		if !page.HasMore {
+			break
+		}
+
+		lastID = page.LastId
+	}
```
ViktorT-11, line 137: I think you need to test that each page returns exactly the events you expect on that page, else in the first `GetChannelEvents` call here could in theory return all events and this test would still pass.

chanevents/store.go
```diff
@@ -185,6 +188,8 @@ func (s *Store) GetChannel(ctx context.Context, channelPoint string) (*Channel,
 func (s *Store) AddChannelEvent(ctx context.Context,
 	event *ChannelEvent) error {
 
+	log.Tracef("Adding channel event: %+v", event)
```
ViktorT-11, line 191: Haven't seen us use the "+" option in "%+v" for logging in any other repo then here in Faraday (lnd only has for `t.Fatalf`). But it seems to be filling a function, so cool to learn about it while reviewing this :rocket:.

ViktorT-11 (commented):
Generally this PR looks great, so mainly leaving a question :fire:.

Additionally, you should add details to commit messages before this is merged.

itest/channel_events_test.go
bitromortac: Added some cross-checks.

db/sqlc/queries/chanevents.sql
bitromortac: Good point. Previously I had time-based pagination, but that didn't work well because of timestamp precision. I didn't really consider offset pagination, but a good point that this would be more consistent with other projects and doesn't expose internal state. However, I see that in the future we may want to delete older events, because I expect a constant stream of updates. In a new terminal web feature I think we also want to sync balances. With deletion I guess that it's easier to maintain the sync state by using the id. Let me know if you prefer the offset way, though I think the id approach has also speed advantages.

ViktorT-11: Ok, ultimately I'll let you decide on which approach you'd want to use!

Personally I have no strong preference. I do think the `offset` approach is usually standard practise though, even when deletion is possible. I think the main drawback with the current approach is that it'll bind the store to always use an incremental `id` for the rows in the `db`.

But ultimately I think there are pros with the current approach as well, so I'll let you decide :)

itest/channel_events_test.go
ViktorT-11: Thanks for the update :tada:!

ViktorT-11 (approved):
Thanks for the update and responding to my previous questions. I'm good to merge this as is if you prefer the current approach. uTACK LGTM :fire:

db/sqlc/queries/chanevents.sql
bitromortac: Ok, I thought about it a bit more, I'd accept the small implementation leak for the performance and for incremental sync that survives head deletions (which I think isn't possible in a straightforward manner with an offset unless one introduces an additional absolute sync watermark value like a timestamp). Thanks for the input here!

PR #240, opened 2026-05-12
forwarding ability: introduce db helpers

bitromortac:
Introduces helpers for the forwarding ability algorithm. A preview for how this is used can be seen here https://github.com/bitromortac/faraday/tree/2604-fwd-prep-11.

### DIFF
--- a/chanevents/quantile.go
+++ b/chanevents/quantile.go
@@ -1,0 +1,55 @@
+package chanevents
+
+import (
+	"errors"
+	"sort"
+
+	"golang.org/x/exp/constraints"
+)
+
+// number is the type constraint Quantile accepts: any sortable numeric type.
+type number interface {
+	constraints.Integer | constraints.Float
+}
+
+// Quantile computes the q-quantile of a slice of comparable values. This can be
+// used to compute the median (q=0.5) or the min (q=0) or max (q=1).
+func Quantile[T number](xs []T, q float64) (float64, error) {
+	if q < 0 || q > 1 {
+		return 0, errors.New("quantile must be between 0 and 1")
+	}
+
+	if len(xs) == 0 {
+		return 0, errors.New("cannot compute quantile of empty slice")
+	}
+
+	if len(xs) == 1 {
+		return float64(xs[0]), nil
+	}
+
+	// Create a copy of the slice to avoid mutating the original.
+	ys := make([]T, len(xs))
+	copy(ys, xs)
+
+	sort.Slice(ys, func(i, j int) bool {
+		return ys[i] < ys[j]
+	})
+
+	// Compute fractional index of q-quantile.
+	if q == 1.0 {
+		return float64(ys[len(ys)-1]), nil
+	}
+	i := q * float64(len(ys)-1)
+
+	// Interpolate between the two consecutive values, depending on the
+	// fractional index position in between.
+	lowerIdx := int(i)
+	upperIdx := lowerIdx + 1
+
+	lowerVal := float64(ys[lowerIdx])
+	upperVal := float64(ys[upperIdx])
+
+	indexDiff := i - float64(lowerIdx)
+
+	return lowerVal + (upperVal-lowerVal)*indexDiff, nil
+}
--- a/chanevents/store.go
+++ b/chanevents/store.go
@@ -192,6 +192,30 @@ func (s *Store) GetChannel(ctx context.Context, channelPoint string) (*Channel,
 	}, nil
 }
 
+// GetChannelByShortChanID retrieves a channel by its short channel ID,
+// returning ErrUnknownChannel if no row matches.
+func (s *Store) GetChannelByShortChanID(ctx context.Context,
+	shortChannelID uint64) (*Channel, error) {
+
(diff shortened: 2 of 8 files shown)


chanevents/store.go
```diff
@@ -22,6 +23,11 @@ var (
 	ErrUnknownChannel = errors.New("unknown channel")
 )
 
+// channelEventsIterChunkSize bounds the per-fetch row count of
+// GetChannelEventsIter so the iterator caps memory at one chunk while streaming
+// long ranges.
+const channelEventsIterChunkSize = 1000
```
ViktorT-11: Should there potentially optionally be an alternative to specify the used value of the `IterChuckSize` through the API?

chanevents/store_test.go
```diff
@@ -204,3 +242,144 @@ func TestPagination(t *testing.T) {
...
+
+	startTime := baseTime
+	endTime := baseTime.Add(2 * time.Hour)
+
+	// Drive the iterator. After consuming the first chunk, inject a
+	// concurrent insert with a timestamp later than the seed events but
+	// still within the iterator's window. This event should be included
+	// since it falls within [startTime, endTime).
+	var iterEvents []*ChannelEvent
+	for event, err := range store.GetChannelEventsIter(
```
ViktorT-11: For additional coverage we could potentially add additional test coverage (doesn't have to be in this specific test) that shows that if the `for` loop reaches a case that triggers a `break` prior to consuming all events, not all events are contained in the `iterEvents` array.

I.e. if it breaks at `len(iterEvents) == 1000` for this specific store, the `iterEvents` array would only contain 1000 events.

chanevents/store_test.go
```diff
@@ -204,3 +242,144 @@ func TestPagination(t *testing.T) {
...
+		ctx, startTime, endTime, channelID,
+	) {
+		require.NoError(t, err)
+		iterEvents = append(iterEvents, event)
+
+		if len(iterEvents) == 1000 {
+			clock.SetTime(baseTime.Add(2*time.Hour - time.Minute))
```
ViktorT-11: nit: To make it really clear and easy to spot that this is after the last seeded event, I would potentially change this to:
`baseTime.Add(time.Duration(seedcount+2) * time.Second),`

chanevents/store.go
```diff
@@ -257,6 +295,53 @@ func (s *Store) GetChannelEvents(ctx context.Context, channelID, afterID int64,
...
+		}
+		scidToPeer[uint64(dbChannel.ShortChannelID)] = dbChannel.Pubkey
+	}
+
+	return scidToPeer, nil
+}
+
+// GetLatestChannelUpdateBefore returns the latest channel event before a given
+// time (exclusive). If no event is found, it returns (nil, nil).
+func (s *Store) GetLatestChannelUpdateBefore(ctx context.Context,
```
ViktorT-11, line 317: Could potentially make it optional to specify which event type this function uses (and then rename the function)?

chanevents/store.go
```diff
@@ -257,6 +295,53 @@ func (s *Store) GetChannelEvents(ctx context.Context, channelID, afterID int64,
...
+	}
+
+	scidToPeer := make(map[uint64]string, len(dbChannels))
+	for _, dbChannel := range dbChannels {
+		// The short channel ID can be zero if it's not known yet. We
+		// should just ignore those.
+		if dbChannel.ShortChannelID == 0 {
+			continue
+		}
+		scidToPeer[uint64(dbChannel.ShortChannelID)] = dbChannel.Pubkey
```
ViktorT-11, line 309: nit: New line after:
```
if dbChannel.ShortChannelID == 0 {
	continue
}
```

ViktorT-11 (approved):
Looks great, LGTM :fire:! Just leaving some suggestions below, non of which are required.

chanevents/store.go
bitromortac: I skipped this as there's only a single caller right now.

chanevents/store.go
bitromortac: The iterator comments are obsolete, I removed it. Thank you for the review!

chanevents/quantile.go
```diff
@@ -0,0 +1,59 @@
...
+
+	lowerVal := ys[lowerIdx]
+	upperVal := ys[upperIdx]
+
+	// Both indexDiff (value between 0 and 1) and valueDiff are always
+	// positive.
+	indexDiff := i - float64(lowerIdx)
+	valueDiff := float64(upperVal - lowerVal)
+
+	return lowerVal + T(valueDiff*indexDiff), nil
```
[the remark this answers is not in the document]
bitromortac: I fixed this by returning float.

db/sqlc/queries/chanevents.sql
```diff
@@ -27,3 +27,14 @@ WHERE channel_id = $1
   AND timestamp < $4
 ORDER BY id ASC
 LIMIT $5;
+
+-- name: GetLatestChannelEventBefore :one
+SELECT * FROM channel_events
```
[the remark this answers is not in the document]
bitromortac: We have that pattern elsewhere as well (won't fix).

PR #241, opened 2026-05-15
forwarding ability: add main algo

bitromortac:
Implements a pair-wise uptime and forwarding velocity algorithm.

### DIFF
--- a/chanevents/analyzer.go
+++ b/chanevents/analyzer.go
@@ -7,6 +7,7 @@ import (
 	"iter"
 	"log/slog"
 	"math"
+	"sort"
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
@@ -80,6 +81,14 @@ type ForwardingAbility struct {
 	Inconsistent bool
 }
 
+// PeerPair identifies a unidirectional routing edge from PeerIn to PeerOut.
+// PeerIn names the source-side peer (the incoming channel's far end in lnd's
+// forwarding vocabulary) and PeerOut names the sink-side peer.
+type PeerPair struct {
+	PeerIn  string
+	PeerOut string
+}
+
 // pairInputs encapsulates the routing performance thresholds for a single
 // direction.
 type pairInputs struct {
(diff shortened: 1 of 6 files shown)


frdrpcserver/forwarding_ability.go
```diff
@@ -0,0 +1,93 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ForwardingAbility returns per-peer-pair velocity and uptime fraction over the
+// requested window. Invalid inputs produce typed gRPC error codes
+// (InvalidArgument, FailedPrecondition).
+func (s *RPCServer) ForwardingAbility(ctx context.Context,
+	req *frdrpc.ForwardingAbilityRequest) (
+	*frdrpc.ForwardingAbilityResponse, error) {
```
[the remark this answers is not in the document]
bitromortac: RPC work was removed here. This will be addressed in a follow-up.

chanevents/analyzer.go
```diff
@@ -0,0 +1,878 @@
...
+				sumBLocal -= state.localBalance
+			}
+			if err := applyEvent(state, event); err != nil {
+				return nil, nil, err
+			}
+			if state.online {
+				sumBRemote += state.remoteBalance
+				sumBLocal += state.localBalance
+			}
+		}
```
ViktorT-11, line 775: For readability, I think it's worth adding a comment of why the state's remote & local balance is first deducted, and then re-applied after the event has been applied.

chanevents/analyzer.go
```diff
@@ -0,0 +1,878 @@
...
+			sumARemote += s.remoteBalance
+			sumALocal += s.localBalance
+		}
+	}
+	for _, s := range statesB {
+		if s.online {
+			sumBRemote += s.remoteBalance
+			sumBLocal += s.localBalance
+		}
+	}
```
ViktorT-11: Nit: for those with trace on, this can potentially also already be done above, and skipped here.

chanevents/analyzer_test.go
```diff
@@ -0,0 +1,904 @@
...
+	err = store.AddChannelEvent(
+		ctx, &ChannelEvent{
+			ChannelID:     channelID,
+			EventType:     EventTypeUpdate,
+			Timestamp:     sameTime,
+			LocalBalance:  fn.Some(btcutil.Amount(200)),
+			RemoteBalance: fn.Some(btcutil.Amount(800)),
+		},
+	)
+	require.NoError(t, err)
```
ViktorT-11, line 786: It would also be good if you could add test coverage which shows that events added for the `sameTime` which has an even higher ID, but which are not `update` events do still lead to that the `update` event with the highest ID gets picked.

chanevents/analyzer_test.go
```diff
@@ -0,0 +1,904 @@
...
+			},
+		},
+	}
+
+	a := NewForwardingAnalyzer(store, lndclient.LndServices{Client: stub})
+
+	startTime := seedTime.Add(time.Second)
+	endTime := startTime.Add(time.Minute)
+
+	abilities, err := a.EffectiveUptime(ctx, startTime, endTime, 0, 0)
```
ViktorT-11, line 937: I think you should also add test coverage which tests the `fwdPercentile` & `threshold` args for the `EffectiveUptime` function here as well. It'd also be valuable to add more test coverage for the `startTime` and `endTime`. Especially since these args will be utilised by the Autopilot server if understand correctly.

ViktorT-11 (commented):
Looks great :fire:, I don't really have a lot of valuable feedback except additional test coverage.

I'm also adding a patch file I spoke about earlier, which contains a suggestion for how extra cacheing could be done. Note that the patch is very much a draft, and doesn't have proper formatting etc:
[extra_cache.patch](https://github.com/user-attachments/files/28285921/extra_cache.patch)

From our offline discussions, I believe that the extra caching will not add too much though, given how this functionality is likely to be used. So feel free to ignore it if you don't really think it adds any value.

Other than that, the PR looks good enough to merge from my point of view. But given our offline discussions, you're still not certain about the API format. So feel free to change that if needed, and I'll be ready to re-review that change.

chanevents/analyzer.go
```diff
@@ -0,0 +1,862 @@
...
+
+	for i, peerA := range peers {
+		statesA := initialStates[peerA]
+		sliceA, err := loadPeer(peerA)
+		if err != nil {
+			return nil, err
+		}
+
+		for j := i; j < len(peers); j++ {
+			peerB := peers[j]
```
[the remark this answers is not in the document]
bitromortac: added a context cancellation check

chanevents/analyzer.go
```diff
@@ -0,0 +1,862 @@
...
+			sumARemote += s.remoteBalance
+			sumALocal += s.localBalance
+		}
+	}
+	for _, s := range statesB {
+		if s.online {
+			sumBRemote += s.remoteBalance
+			sumBLocal += s.localBalance
+		}
+	}
```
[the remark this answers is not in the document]
bitromortac: extracted the computation

frdrpcserver/forwarding_ability.go
```diff
@@ -0,0 +1,93 @@
...
+	if req.ForwardPercentile < 0 || req.ForwardPercentile > 100 {
+		return nil, status.Error(
+			codes.InvalidArgument,
+			"forward_percentile must be in [0, 100]",
+		)
+	}
+
+	forwardingAnalyzer := chanevents.NewForwardingAnalyzer(
+		s.cfg.ChanEvents, s.cfg.Lnd,
+	)
```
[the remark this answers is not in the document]
bitromortac: not part of this PR anymore

chanevents/analyzer.go
bitromortac: Yes, I agree, I should have included that right from the start

chanevents/analyzer.go
bitromortac: I think this is outdated now, as I moved the sum calculation to outside of the function

chanevents/analyzer_test.go
bitromortac: Added a case in `TestInitialStateSameSecond`.

chanevents/analyzer_test.go
bitromortac: Added a bit more coverage, but I think the start/end time is probably well covered by other lower-level tests.

chanevents/analyzer.go
```diff
@@ -0,0 +1,904 @@
...
+		if traceOn {
+			log.TraceS(
+				ctx, "Processing event",
+				slog.Int64("chanID", event.ChannelID),
+				btclog.Fmt("type", "%v", event.EventType),
+				slog.Time("time", event.Timestamp),
+			)
+		}
+
+		accumulate(event.Timestamp.Sub(lastTimestamp))
```
ViktorT-11, line 735: nit: maybe also add a comment to explain why the accumulate function is called before the event has been applied (in combination with the accumulate call outside of this for loop.)

ViktorT-11 (approved):
Nice, thanks for the updates. Feels like the PR and algo in general is easier to understand with the new changes. LGTM :fire:

chanevents/analyzer.go
bitromortac: Nice, thank you for the review, I added some comments to `accumulate` calls.

PR #245, opened 2026-06-23
chanevents: add channel event pruning

bitromortac:
Bounds the `channel_events` table, which otherwise grows without bound on high-frequency channels. Pruning is on by default with two independent limits:

- `--chanevents.max-events=7000000` (~1GB of storage)
- `--chanevents.retention=1440h (60 days)`

Both parameters can be used independently and the pruning interval is about an hour.

### DIFF
--- a/chanevents/monitor.go
+++ b/chanevents/monitor.go
@@ -18,6 +18,16 @@ const (
 	// retryInterval is the time to wait before retrying after a
 	// transient error or while waiting for lnd to become ready.
 	retryInterval = 5 * time.Second
+
+	// pruneInterval is how often the monitor enforces the channel event
+	// storage limits while consuming live events.
+	pruneInterval = time.Hour
+
+	// minPruneInterval is the floor for the background pruning ticker. A
+	// tiny retention window would otherwise drive the ticker interval down
+	// to milliseconds and starve the CPU, so we never tick faster than
+	// this.
+	minPruneInterval = time.Second
 )
 
 var (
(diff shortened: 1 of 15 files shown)


db/sqlc/queries/chanevents.sql
```diff
@@ -38,3 +38,19 @@ LIMIT 1;
...
+-- max-events rows by deleting everything with a smaller (earlier-inserted) id
+-- than the max-events-th newest id. The age limit deletes any row whose
+-- timestamp predates timestamp_cutoff.
+DELETE FROM channel_events
+WHERE (sqlc.arg(use_size_limit) = 1 AND channel_events.id < COALESCE((
+    SELECT id FROM channel_events
+    ORDER BY id DESC
+    LIMIT 1 OFFSET sqlc.arg(offset_count)
+), 0))
+OR (sqlc.arg(use_age_limit) = 1 AND channel_events.timestamp < sqlc.arg(timestamp_cutoff));
```
ViktorT-11: IMO, for readability, it's probably better here to sacrifice a little performance to instead separate thi into 2 different queries i.e. one separate query for deleting by size limit, and one for age limit.

Then the `store` will decide if both or only one of the queries gets executed depending on if deletion by size or age is activated.

That way we could probably also remove the need for the `sqlc.arg` params, which we've tried to avoid using in our repos.

itest/channel_events_test.go
```diff
@@ -360,3 +360,68 @@ func TestForwardingDowntime(t *testing.T) {
...
+	)
+	require.NoError(c.t, err, "could not get channel events")
+	require.Greater(
+		c.t, len(eventsBefore.Events), 1,
+		"expected multiple events to be recorded initially",
+	)
+
+	// Wait for the live background pruning ticker to bound the table to the
+	// max-events ceiling. We assert at most one event rather than exactly
+	// one: the size limit keeps a single event, but the 2-second retention
```
ViktorT-11, line 400: Potentially, you could also add a check after this that checks that the events finally go down to 0 once all events are too old to be kept.

chanevents/monitor.go
ViktorT-11, line 249: We could maybe also activate pruning once a number of events of a threshold limit has been handled, which resets the ticker. Just to avoid that we'd get a lot of events in one hour which makes the db much larger than the intended max size.

ViktorT-11 (approved):
Looks good! Just leaving a few suggestions which are non-blocking

db/sqlc/queries/chanevents.sql
bitromortac: I agree, split it into two parts :+1:.

itest/channel_events_test.go
bitromortac: Done, good way to test the retention further!

chanevents/monitor.go
bitromortac: I didn't address this, I think we can get to that later if we really need such high-fidelity

PR #248, opened 2026-08-18
mod: bump go version and sqldb version

bitromortac:
This makes faraday compatible with the latest lit master branch.


ViktorT-11 (approved):
LGTM!

PR #251, opened 2026-08-26
forwarding ability: change forwarding volume to msat in RPC

bitromortac:
Changes the `ForwardingAbility` endpoint to report the volume in msat to prevent precision loss. The effective uptime is changed to a uint64 in order to be consistent in the handling of durations in the proto.


frdrpcserver/forwarding_ability_test.go
```diff
@@ -172,7 +173,7 @@ func TestForwardingAbility(t *testing.T) {
 				return map[chanevents.PeerPair]chanevents.ForwardingAbility{
 					{PeerIn: peerIn, PeerOut: peerOut}: {
 						EffectiveUptime: 10 * time.Second,
-						ForwardedAmount: 999,
+						ForwardedMsat:   999,
```
ViktorT-11: nit: should be updated to `999_000` as it's now in msat.

frdrpc/forwarding_ability_codec_test.go
```diff
@@ -51,15 +51,15 @@ func TestForwardingAbilityCodecRoundTrip(t *testing.T) {
 				fwdKey(1): {
 					fwdKey(2): {
 						EffectiveUptimeS: 80,
-						ForwardedSat:     1500,
+						ForwardedMsat:    1500,
```
ViktorT-11: nit: should probably also multiply the values by 1000 for these updates in this test file, just for consistency.

ViktorT-11 (approved):
LGTM :fire:! Just leaving 2 nits leaving some consistency feedback.

PR #253, opened 2026-09-02
config: support reading bitcoin rpc password from a file

### LINKED ISSUE #252: Can you support reading password from a file?
Contacting you as lightning terminal will stop working properly in btcpayserver docker-deployment.

We are working to remove any hardcoded password from our configs, to harden security but faraday rely on one as you can [see on file](https://github.com/btcpayserver/btcpayserver-docker/blob/master/docker-compose-generator/docker-fragments/opt-add-lightning-terminal.yml#L33)

Ideally, I would like to change the line like this.
```diff
- - "--faraday.bitcoin.password=afixedpasswordbecauselndsuckswithcookiefile"
+ - "--faraday.bitcoin.passwordfile=/path/pwd-file"
```

If you can make a new version with this feature, I can update the lightning terminal fragment in our repo.

Roasbeef:
## Summary

Adds a `--bitcoin.passwordfile` config option as an alternative to
`--bitcoin.password`, so operators can point faraday at a file
containing the bitcoind/btcd RPC password instead of putting it in
plaintext on the command line or in a docker-compose manifest.

- `chain.BitcoinConfig` gains a `PasswordFile` field
  (`--bitcoin.passwordfile`).
- `ValidateConfig` resolves the password from the file when
  `--bitcoin.passwordfile` is set, trimming surrounding whitespace and
  the trailing newline, and rejects the config if both
  `--bitcoin.password` and `--bitcoin.passwordfile` are set, if the
  file can't be read, or if it's empty.
- README documents the new flag.

Closes #252.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race .` (new `TestValidateConfigBitcoinPassword` covers
      the plaintext flag, the file flag, a missing file, an empty
      file, and the mutually-exclusive-flags case)
- [x] `go test ./...` (itest failures are pre-existing/unrelated —
      that suite requires a live bitcoind/lnd docker environment)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,14 @@ To connect Faraday to btcd:
 --bitcoin.tlspath={path to btcd cert}
 ```
 
+Instead of passing the password on the command line with `--bitcoin.password`,
+it can be read from a file with `--bitcoin.passwordfile`, which keeps the
+password out of the process list and shell history:
+```text
+--bitcoin.passwordfile={path to a file containing the rpc password}
+```
+`--bitcoin.password` and `--bitcoin.passwordfile` are mutually exclusive.
+
 #### RPCServer
 Faraday serves requests over grpc by default on `localhost:8465`. This default can be overwritten:
 ```text
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ To connect Faraday to btcd:
 
 Instead of passing the password on the command line with `--bitcoin.password`,
 it can be read from a file with `--bitcoin.passwordfile`, which keeps the
-password out of the process list and shell history:
+password out of the process list and shell history; surrounding whitespace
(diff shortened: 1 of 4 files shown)


bitromortac (approved):
LGTM tACK :tada:

NicolasDorier:
@Roasbeef I wished I'd get that for LND too. Because LND didn't handle change of cookie value at run time, I needed to configure a fixed password.

I will handle that in our `docker-entrypoint.sh` of our own image for now, but having it in LND would be very nice. (either that, or having your RPC client re-fetch automatically the `__cookie__` pwd when the auth fails)

ViktorT-11 (approved):
Thanks, LGTM 🙏

